Issue #21, opened 2019-09-05
ActiveRecordConsumer assumes key is id when deleting or updating


dorner:
Currently, ActiveRecordConsumer will delete or update a record by finding it based on the key. This assumes that the key is always the primary key of the record, which isn't always true. We should refactor this into a `fetch_record` method which can be overridden similar to other methods.

Even better, use the key schema to fetch records.

PR #32, opened 2019-09-26
Single consumer

colinmroberts:
Combining the consumers and batch consumers into a single file. This still has some issues to be worked out, but this is the general approach


lib/deimos/activerecord/batch_consumption.rb
```diff
@@ -0,0 +1,139 @@
...
+      # @return [Hash] The key attributes.
+      def record_key(key)
+        decoded_key = decode_key(key)
+
+        if decoded_key.nil?
+          {}
+        elsif decoded_key.is_a?(Hash)
+          @key_converter.convert(decoded_key)
+        else
+          { @klass.primary_key => decoded_key }
```
dorner: FYI #21 would still need to be fixed with this. I think it might be too much work to be bundled in here though.

lib/deimos/activerecord/batch_consumption.rb
```diff
@@ -0,0 +1,139 @@
...
+      # records. Used for bulk deletion.
+      # @param records [Array<Array>] List of [key, nil] pairs.
+      # @return ActiveRecord::Relation Matching relation.
+      def deleted_query(records)
+        keys = records.
+          map { |k, _| record_key(k) }.
+          reject(&:empty?)
+
+        keys.reduce(@klass.none) do |query, key|
+          query.or(@klass.unscoped.where(key))
```
dorner: Can we not use `or`? This will crash on older versions of Rails. If it's a single field (and I think we can assume that case, allowing people to override this query where it isn't) you can just use `@klass.unscoped.where(key_field: decoded_keys)`. It would mean revamping how `record_key` works but I think that's probably better in any case - the keys of that query (the fields to check) should be identical for all records in the same topic.

lib/deimos/activerecord/batch_consumption.rb
```diff
@@ -0,0 +1,139 @@
...
+
+      # Get the set of attribute names that uniquely identify messages in the
+      # batch. Requires at least one record.
+      # @param records [Array<Array>] Non-empty list of [key, value] pairs.
+      # @return [Array<String>] List of attribute names.
+      # @raise If records is empty.
+      def key_columns(records)
+        raise 'Cannot determine key from empty batch' if records.empty?
+
+        first_key, = records.first
```
dorner: Why the , ?

lib/deimos/activerecord/batch_consumption.rb
```diff
@@ -0,0 +1,139 @@
...
+        end
+      end
+
+      # Get the set of attribute names that uniquely identify messages in the
+      # batch. Requires at least one record.
+      # @param records [Array<Array>] Non-empty list of [key, value] pairs.
+      # @return [Array<String>] List of attribute names.
+      # @raise If records is empty.
+      def key_columns(records)
+        raise 'Cannot determine key from empty batch' if records.empty?
```
dorner: Should this really crash?

lib/deimos/activerecord/batch_consumption.rb
```diff
@@ -0,0 +1,139 @@
...
+        end
+
+        ops = messages.group_by { |k, _| k }
+
+        # Find maximum depth
+        depth = ops.values.map(&:length).max || 0
+
+        # Generate slices for each depth
+        depth.times.map do |i|
+          ops.values.map { |arr| arr.dig(i) }.compact
```
dorner: See above comment - is there a reason we even try to insert the a message with the same key but different values? By definition the last value will "win" so we can entirely ignore the earlier values.

lib/deimos/activerecord/message_consumption.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
```
dorner: The folder name should be `active_record` to be consistent with CamelCase -> snake_case conversion.

lib/deimos/consume/batch_consumption.rb
```diff
@@ -0,0 +1,143 @@
+# frozen_string_literal: true
```
dorner: Shouldn't this be in a `consumer` folder instead of a `consume` folder?

lib/deimos/consume/message_consumption.rb
```diff
@@ -0,0 +1,91 @@
...
+        Deimos.config.logger.warn(
+          message: 'Error consuming message',
+          handler: self.class.name,
+          metadata: metadata,
+          data: payload,
+          error_message: exception.message,
+          error: exception.backtrace
+        )
+        super
+      end
```
dorner: Wouldn't this clobber the `_handle_error` of the batch consumption module if both are included?

lib/deimos/consumer.rb
```diff
@@ -1,99 +1,106 @@
...
-        payload: payload,
-        time_elapsed: time_taken,
-        metadata: metadata
-      )
+    def _handle_error(exception, payload, metadata)
+      Deimos.config.tracer&.set_error(@span, exception)
+
+      raise if Deimos.config.reraise_consumer_errors ||
+               Deimos.config.fatal_error_block.call(exception, payload, metadata) ||
+               fatal_error?(exception, payload, metadata)
```
dorner: And wouldn't this clobber both of them?

lib/deimos/active_record_consumer.rb
```diff
@@ -1,53 +1,64 @@
...
-        destroy_record(record)
-        return
-      end
-      record ||= @klass.new
-      attrs = record_attributes(payload.with_indifferent_access)
-      # don't use attributes= - bypass Rails < 5 attr_protected
-      attrs.each do |k, v|
-        record.send("#{k}=", v)
+      if self.class.config[:key_schema] # rubocop:disable Style/GuardClause
+        @key_converter = SchemaModelConverter.new(
```
dorner: This... doesn't exist?

lib/deimos/active_record_consumer.rb
```diff
@@ -1,53 +1,64 @@
...
+  protected
+
+    # Get attributes for new/upserted records in the database. Override this
+    # method (with super) to customize the set of attributes used to instantiate
+    # records.
+    # @param payload [Hash] The decoded message payload.
+    # @param key [Hash] The decoded message key.
+    # @return [Hash|nil] Attribute set for the upserted record. nil to skip
+    # insertion.
+    def record_attributes(key, payload)
```
dorner: Can we change this to `record_attributes(payload, key=nil)` to preserve backwards compatibility?

PR #35, opened 2019-11-05
New config

### LINKED ISSUE #18: Add DDL for configuration
Currently configuration is split up between the producer/consumer class, a `configure` app-wide block, and `phobos.yml`. We should strike our own way for this and add a DDL similar to Rails routes where we can specify overall configuration and producer/consumer configs in a single place.

Truth is, even if we can combine the `configure` block and the .yml file together that would be a win. I'm not 100% sure if the producer/consumer config should live with the individual class or centralized.

dorner:
# Pull Request Template

## Description

Fixes #18. Backwards compatibility should be completely transparent due to deprecation capability.

This introduces essentially a brand new configuration framework. We can consider breaking this out into its own gem which would be neat since it does things that (AFAICT) no other configuration gem does.

## Type of change

## How Has This Been Tested?

Local testing with existing Deimos app as well as new unit tests.

## Checklist:


lib/deimos/config/phobos_config.rb
```diff
@@ -0,0 +1,132 @@
...
+      }
+
+      p_config[:listeners] = self.consumer_objects.map do |consumer|
+        hash = consumer.to_h.reject do |k, _|
+          %i(class_name schema namespace key_config backoff).include?(k)
+        end
+        hash = hash.map { |k, v| [k, v.is_a?(Symbol) ? v.to_s : v] }.to_h
+        hash[:handler] = consumer.class_name
+        if consumer.backoff
+          hash[:backoff] = {
```
colinmroberts: Maybe extract out this Range -> backoff hash creation, since it's done in a couple of places

lib/deimos/config/phobos_config.rb
```diff
@@ -0,0 +1,132 @@
...
+          p_config[:kafka]["ssl_#{key}".to_sym] = ssl_var_contents(self.kafka.ssl.send(key))
+        end
+      end
+      p_config
+    end
+
+    # @param key [String]
+    # @return [String]
+    def ssl_var_contents(key)
+      File.exist?(key) ? File.read(key) : key
```
colinmroberts: Feels a little iffy mixing certs and file paths like this. I feel like they should be more explicitly separated, e.g. with a `file://` scheme or something. This feature wasn't supported in the old configs right?

lib/deimos/config/configuration.rb
```diff
@@ -0,0 +1,330 @@
...
+      # :all to log all topics.
+      setting :log_topics, []
+
+      # @return [Symbol|Array<String>] A list of topics to compact messages for
+      # before sending, or :all to compact all keyed messages.
+      setting :compact_topics, []
+
+    end
+
+    setting_object :producer do
```
colinmroberts: TODO: Docs for these

lib/deimos/config/configurable.rb
```diff
@@ -0,0 +1,258 @@
...
+        else
+          self.value = self.default_value
+        end
+      end
+
+      # :nodoc:
+      def clone
+        setting = ConfigSetting.new(self.value, self.default_value,
+                                    self.default_proc, self.deprecation)
+        setting.reset!
```
colinmroberts: Just for clarity, why does cloning reset the new object?

lib/deimos/config/configurable.rb
```diff
@@ -0,0 +1,258 @@
...
+      #   setting_object :producer do
+      #     setting :topic
+      #     setting :class_name
+      #   end
+      # This will create the `producer` method to define these values as well
+      # as the `producer_objects` method to retrieve them.
+      # @param name [Symbol]
+      def setting_object(name, &block)
+        new_config = ConfigStruct.new("#{@name}.#{name}")
+        @setting_objects[name] = []
```
colinmroberts: May be good to add a warning if name is redefined

lib/deimos/config/configurable.rb
```diff
@@ -0,0 +1,258 @@
...
+      # @param name [Symbol]
+      # @param default_value [Object]
+      # @param default_proc [Proc]
+      def setting(name, default_value=nil, default_proc: nil, &block)
+        if block_given?
+          # Create a nested setting
+          setting_config = @settings[name]&.value || ConfigStruct.new("#{@name}.#{name}")
+          setting = ConfigSetting.new
+          setting.value = setting_config
+          @settings[name] = setting
```
colinmroberts: I wonder if it would be useful to add a warning somehow if you accidentally redefine the same name inside a `configure` block

lib/deimos/config/configurable.rb
```diff
@@ -0,0 +1,258 @@
...
+      # Define a new setting object and use the passed block to define values.
+      def _new_setting_object_method(config_key, &block)
+        new_config = @setting_templates[config_key].clone
+        new_config.instance_eval(&block)
+        @setting_objects[config_key] << new_config
+      end
+
+      # Return a setting object.
+      def _setting_object_method(config_key)
+        key = config_key.to_s.sub('_objects', '').to_sym
```
colinmroberts: Should add an $ anchor to the regex

lib/deimos/config/configuration.rb
```diff
@@ -0,0 +1,330 @@
...
+    include Deimos::PhobosConfig
+  end
+
+  # :nodoc:
+  after_configure do
+    Phobos.configure(self.config.phobos_config)
+    self.config.producer_objects.each do |producer|
+      configure_producer_or_consumer(producer)
+    end
+    self.config.consumer_objects.each do |consumer|
```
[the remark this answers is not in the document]
dorner: It shouldn't be, since we've moved everything to a single config file. I'm not sure what the goal would be of having them. If the issue is that there's a consumer with a topic that isn't encoded, one of the things on the roadmap would be to specify different backends for topics (e.g. using Avro without a schema registry, JSON schema, no encoding at all, Protobuf etc.) which would allow this behavior.

lib/deimos/config/phobos_config.rb
dorner: It was, actually. :) Truth is I can deprecate these since RubyKafka itself has separate configs for files or strings. Always felt a bit weird.

lib/deimos/config/configurable.rb
dorner: Hmm - you're right, this shouldn't be a clone method. I'll rename it.

lib/deimos/config/configurable.rb
dorner: Might be tricky to detect... I think it's overkill for now.

dorner:
Thanks for the comments @colinmroberts! Will address this week hopefully.

docs/CONFIGURATION.md
```diff
@@ -0,0 +1,178 @@
...
+kafka.seed_brokers|`['localhost:9092']`|URL for the Kafka brokers.
+kafka.client_id|`phobos`|Identifier for this application.
+kafka.connect_timeout|15|The socket timeout for connecting to the broker, in seconds.
+kafka.socket_timeout|15|The socket timeout for reading and writing to the broker, in seconds.
+kafka.ssl.enabled|false|Whether SSL is enabled on the brokers.
+kafka.ssl.ca_cert|nil| A PEM encoded CA cert, a file path to the cert, or an Array of certs to use with an SSL connection.
+kafka.ssl.client_cert|nil|A PEM encoded client cert to use with an SSL connection, or a file path to the cert.
+kafka.ssl.client_cert_key|nil|A PEM encoded client cert key to use with an SSL connection.
+
+## Consumer Configuration
```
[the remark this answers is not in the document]
dorner: Actually only consumers allow you to override, not producers. Can always change that if there's demand.

spec/config/configurable_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+    MyConfig.configure do
+      setting :group do
+        setting :set6, 20
+      end
+    end
+    expect(MyConfig.config.group.set6).to eq(20)
+    expect(MyConfig.config.group.set5.call).to eq(15)
+
+  end
+
```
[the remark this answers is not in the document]
dorner: There's a separate spec for deprecated properties in `configuration_spec.rb`.

lib/deimos/config/phobos_config.rb
dorner: Hm went back and checked and it only supports file path for ca_cert, not for the other options. Let's leave it as is for now.

lib/deimos/config/configurable.rb
dorner: I think this might get noisy. Extending settings is supported fully by the framework so I don't see this as a warning.

dorner:
@colinmroberts addressed the comments; can you give it another look?

PR #42, opened 2019-12-16
Change KafkaSource destroy callback to immediately publish instead of using send_events

### LINKED ISSUE #39: Better way to handle deletes in `generate_payload`
Currently `generate_payload` is called even for deletion events via KafkaSource. This forces implementing code to check to see if the only key is the payload key and return otherwise. We should intentionally not call `generate_payload` on deletion events.

csingh:
## Description

Just a quick change so that KafkaSource destroy callback skips the send_event workflow, as its not required for a simple "delete" payload

Fixes #39 

## Type of change

## How Has This Been Tested?

- [x] Ran the RSpecs, [kafka_source_spec.rb](https://github.com/flipp-oss/deimos/blob/master/spec/kafka_source_spec.rb#L87) covers the case where `.destroy` on an ActiveRecord object should result in a nil payload published to the topic

## Checklist:

### DIFF
--- a/lib/deimos/kafka_source.rb
+++ b/lib/deimos/kafka_source.rb
@@ -41,7 +41,7 @@ module Deimos
     def send_kafka_event_on_destroy
       return unless self.class.kafka_config[:delete]
 
-      self.class.kafka_producers.each { |p| p.send_event(self.deletion_payload) }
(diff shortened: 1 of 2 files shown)


csingh:
@dorner added a spec!

dorner:
Looks good! 👍

PR #49, opened 2020-01-14
- Added `db_producer.insert` and `db_producer.process` metrics.

dorner:


spec/backends/db_spec.rb
```diff
@@ -4,6 +4,11 @@
   include_context 'with publish_backend'
 
   it 'should save to the database' do
+    expect(Deimos.config.metrics).to receive(:increment).ordered.with(
```
colinmroberts: Don't think the `.ordered` here does anything

PR #51, opened 2020-01-30
CCAP-59: Added session_id for KafkaMessage batches in db_producer

george-ma:
Created migration to add an (indexed) session_id to the kafka_messages table
Updated db_producer logic to update the session_id when a batch of messages are received (this session_id is used in the DELETES)
Updated .md files (version breaking change)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,63 +7,70 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## UNRELEASED
 
-# [1.6.0] - 2020-01-30
+## [1.6.0] - 2020-01-30
+
 - Added a session_id column (string) to the kafka_messages table.
 - Added an index for this session_id column.
 - DB Producer now updates the session_id column with a UUID when a batch of KafkaMessages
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## UNRELEASED
 
+# [1.6.0] - 2020-01-30
+- Added a session_id column (string) to the kafka_messages table.
+- Added an index for this session_id column.
+- DB Producer now updates the session_id column with a UUID when a batch of KafkaMessages
+  is retrieved, and uses this UUID when the messages are deleted.
```
dorner: Please leave this under UNRELEASED rather than a specific version.

lib/deimos/utils/db_producer.rb
```diff
@@ -79,20 +79,26 @@ def process_topic_batch
         messages = retrieve_messages
         return false if messages.empty?
 
+        # Update session_id column for this batch of KafkaMessages
+        @batch_uuid = SecureRandom.uuid
```
dorner: This doesn't need to be an instance variable.

PR #71, opened 2020-06-14
ActiveRecord Batch Consumer

colinmroberts:
# Pull Request Template

## Description

ActiveRecord Consumer using batch consumption

## Type of change

## How Has This Been Tested?

Added lots of new tests

## Checklist:


README.md
```diff
@@ -555,6 +558,58 @@ class MyConsumer < Deimos::ActiveRecordConsumer
...
+**Note**: Currently, batch consumption only supports only primary keys as identifiers out of the box. See
+[the specs](spec/active_record_batch_consumer_spec.rb) for an example of how to use compound keys.
+
+A sample batch consumer would look as follows:
+
+```ruby
+class MyConsumer < Deimos::ActiveRecordConsumer
+  schema 'MySchema'
+  key_config field: 'my_field'
+  record_class Widget
```
dorner, line 592: These have been moved to the configuration side.

lib/deimos/active_record/batch_consumption.rb
```diff
@@ -0,0 +1,179 @@
...
+      # @param payloads [Array<Hash>] Decoded payloads.
+      # @param metadata [Hash] Information about batch, including keys.
+      def consume_batch(payloads, metadata)
+        # Create [key, value] pairs, even if no keys are decoded
+        messages = payloads.
+          zip(metadata[:keys]).
+          map { |p, k| [k, p] }
+
+        Deimos.instrument('ar_consumer.consume_batch', topic: metadata[:topic], messages: messages) do
+          slices = _slice_batch(messages)
```
dorner, line 17: Can you add some comments as to the overall algorithm here? I'm having trouble following what the point of these slices is.

dorner: I think a lot of the difficulty I have is in the types that are flying around here, including hashes and arrays of arrays etc. Can we bundle some of this up into a data class?

lib/deimos/active_record/batch_consumption.rb
```diff
@@ -0,0 +1,179 @@
...
+      # Upsert any non-deleted records
+      # @param records [Array<Array>] List of [key, value] pairs for a group of
+      # non-tombstone records.
+      def upsert_records(records)
+        key_cols = key_columns(records)
+
+        messages = records.map do |k, v|
+          record_attributes(v, k).
+            merge(record_key(k))
+        end
```
dorner, line 62: These are basically record hashes, not messages, right? These are things that will be imported.

lib/deimos/active_record/batch_consumption.rb
```diff
@@ -0,0 +1,179 @@
...
+      # Retry the given block when encountering a deadlock. For any other
+      # exceptions, they are reraised. This is used to handle cases where
+      # the database may be busy but the transaction would succeed if
+      # retried later.
+      def _retry_deadlock(tags=[])
+        count = RETRY_COUNT
+
+        begin
+          yield
+        rescue ::ActiveRecord::Deadlocked => e
```
dorner, line 159: I know at least in older MySQL you'd get a different error than this for an actual deadlock (it would be a `Mysql2::Error`. Can we confirm that this will work across at least MySQL and Postgres?

lib/deimos/active_record/batch_consumption.rb
```diff
@@ -0,0 +1,179 @@
...
+          raise e if count <= 0
```
dorner, line 160: Can just do `rescue ActiveRecord::Deadlocked` and `raise` if you never reference the `e` variable.

lib/deimos/active_record/message_consumption.rb
```diff
@@ -0,0 +1,57 @@
+# frozen_string_literal: true
+
+module Deimos
+  module ActiveRecord
+    # Consumption methods
+    module MessageConsumption
```
dorner, line 6: Starting to get confused with all these modules and how they relate to the consumers now - can you cleanly state what each module is responsible for? If not, maybe the modules should be split up differently?

README.md
```diff
@@ -555,6 +548,58 @@ class MyConsumer < Deimos::ActiveRecordConsumer
...
+
+  # Controls whether the batch is compacted before consuming.
+  # If true, only the last message for each unique key in a batch will be
+  # processed.
+  # If false, each message will be processed in order.
+  compacted false
```
dorner, line 588: Should this also be added to the configuration side?

lib/deimos/active_record/batch_consumption.rb
```diff
@@ -0,0 +1,133 @@
...
+      # in a single database operation.
+      # If two messages in a batch have the same key, we cannot process them
+      # in the same operation as they would interfere with each other. Thus
+      # they are split
+      # @param payloads [Array<Hash>] Decoded payloads.
+      # @param metadata [Hash] Information about batch, including keys.
+      def consume_batch(payloads, metadata)
+        messages = payloads.
+          zip(metadata[:keys]).
+          map { |p, k| Deimos::Message.new(p, nil, key: k) }
```
dorner, line 23: OK - now I understand what the slicing is supposed to achieve. I'm not convinced this whole slicing thing is actually necessary. If the records have an ID, there is literally no point in us saving them three or four times in a single batch. The last value will "win". The only possible reason we could have for updating it in different slices would be if there was some kind of database trigger that happened on update.

I'd prefer to have compaction turned on by default, and have the user turn it off if and only if the table has no primary key, or they know that the topic is insert-only (i.e. there is no primary key, or the primary key is not in the topic).

README.md
colinmroberts: Would this be in the `consumer` `setting_object`? I considered this but I wasn't sure if it made sense there as it only applied to ActiveRecord batch consumers. I can move it there if you think it belongs, but I figured it was very similar to the `record_class` config

dorner: Hmm... yeah, OK. Let's leave it here for now. If we start accumulating more of these we might want to add them in.

lib/deimos/active_record/batch_consumption.rb
colinmroberts: I'm OK with making compacted the default. Will make those changes

There are other situations where it makes sense to have it not be compacted. The consumer may do some custom handling where they need to add some app logic for every event. It is also useful for when the topic represents something like an audit log and you use soft deletion. If you have an entity that gets created and then destroyed immediately after, you would still want that record to appear in your DB with a deleted flag set. We do something similar to this in the item-hippocampus

dorner: You've already got the work done, so I'm OK leaving it in. Just not super stoked about having to maintain this when the use case seems to be pretty obscure. Is there some way we can separate it out a bit more (e.g. having a different subclass that handles this as opposed to the "paved road" where it's compacted)?

colinmroberts: Do you think a subclass of ActiveRecordConsumer would be clean enough? The other alternative I can think of is another module you include to override `consume_batch` but that seems uglier

dorner: I think you can have two different methods at least, one that handles slicing and one that doesn't, and you call them differently depending on settings. The methods could be in separate modules but that might be overkill.

colinmroberts:
@dorner Made some more suggested changes

colinmroberts:
@dorner Fixed the conflict 🎉

dorner:
YAHOO

PR #85, opened 2020-08-19
GH-75: Handle batch of unique messages in compact_messages

### LINKED ISSUE #75: DB Producer can fail if all messages in batch were unique
In the DB producer, if a compacted topic has a batch with all-unique keys, it will fail when calling `compact_messages` with this error:

```
NoMethodError:
       undefined method `reverse!' for nil:NilClass
```

E.g. consider this sample test which should pass:

```
      it 'should no make any changes when there are no duplicates' do
        Deimos.configure { |c| c.db_producer.compact_topics = :all }
        expect(producer.compact_messages(deduped_batch)).to eq(deduped_batch)
      end
```

This appears to be because compact batch is calling `reverse!` on `uniq!`. `uniq!` will return `nil` when no changes were made to the array, i.e. it was unique already. This method should use `uniq` instead

angad-singh:
# Pull Request Template

## Description

`compact_messages` now uses `uniq` instead of `uniq!` which was causing it to fail on batch of unique messages.

Fixes #75 

## Type of change

## How Has This Been Tested?

- [x] Unit tests
- [x] Testing on local environment with Fadmin by monkey patching and using "Flyers.PageItem" topic (locally)

## Checklist:

### DIFF
--- a/spec/utils/db_producer_spec.rb
+++ b/spec/utils/db_producer_spec.rb
@@ -155,6 +155,10 @@ each_db_config(Deimos::Utils::DbProducer) do
         Deimos.configure { |c| c.db_producer.compact_topics = [] }
       end
 
+      it 'should not raise NoMethodError if all the keys are unique' do
+        Deimos.configure { |c| c.db_producer.compact_topics = %w(my-topic my-topic2) }
+        expect{ producer.compact_messages(deduped_batch) }.to_not raise_error(NoMethodError)
+      end
     end
   end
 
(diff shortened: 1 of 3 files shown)


spec/utils/db_producer_spec.rb
```diff
@@ -155,6 +155,10 @@
         Deimos.configure { |c| c.db_producer.compact_topics = [] }
       end
 
+      it 'should not raise NoMethodError if all the keys are unique' do
+        Deimos.configure { |c| c.db_producer.compact_topics = %w(my-topic my-topic2) }
+        expect { producer.compact_messages(deduped_batch) }.not_to raise_error(NoMethodError)
+      end
```
dorner, line 161: I'd say this is kind of a silly way to name and focus the test - the point is not that it does not raise this specific error - you fixed the thing that would cause this. You just need to make sure that this code runs in this situation, so you could say something like "it should compact messages where all messages are unique" and check the result. Automatically it will also check that it doesn't raise an error, because if it does, the spec will fail. :)

angad-singh: Ah yeah you are right, as always! I will update the spec :)

PR #94, opened 2020-10-27
GUILD-574: Fix KafkaSource crashing on imports when import hooks are …

### LINKED ISSUE #73: KafkaSource crashes on bulk-import if import hooks are disabled
If you have a KafkaSource with import hooks disabled:

```ruby
  def self.kafka_config
    {
      update: true,
      delete: true,
      import: false,
      create: true
    }
  end
```

Doing a bulk import that only updates (no inserts) on that class will fail with the following error:

```
NoMethodError: undefined method `num_inserts' for nil:NilClass
```

This seems to be because the early return in the `import_without_validations_or_callbacks` monkey patch does not return the `results` from `super`

```
        results = super
        return unless self.kafka_config[:import] #### Here
        return if array_of_attributes.empty? #### And here
```

The activerecord-import gem later tries to use the results of the monkey patched method like this and crashes (because it's nil). See import.rb:735

```
      return_obj.num_inserts = 0 if return_obj.num_inserts.nil?
```

DeeChau:
…disabled

# Pull Request Template

## Description

Return `ActiveRecord::Import::Result.new([], 0, [], [])` when the kafka_config `import` is not set or the result set is empty.
Fixes an issue where KafkaSource crashes on bulk-imports if import hooks are disabled.

Fixes #73

## Type of change

## How Has This Been Tested?

- [x] Unit Tests

## Checklist:

### DIFF
--- a/lib/deimos/kafka_source.rb
+++ b/lib/deimos/kafka_source.rb
@@ -89,7 +89,7 @@ module Deimos
                                                   options={})
         results = super
         if !self.kafka_config[:import] || array_of_attributes.empty?
-          return ActiveRecord::Import::Result.new([], 0, [], [])
+          return results
         end
 
         # This will contain an array of hashes, where each hash is the actual
--- a/lib/deimos/kafka_source.rb
+++ b/lib/deimos/kafka_source.rb
@@ -88,8 +88,9 @@ module Deimos
                                                   array_of_attributes,
                                                   options={})
         results = super
-        return unless self.kafka_config[:import]
-        return if array_of_attributes.empty?
+        if !self.kafka_config[:import] || array_of_attributes.empty?
+          return ActiveRecord::Import::Result.new([], 0, [], [])
+        end
 
         # This will contain an array of hashes, where each hash is the actual
         # attribute hash that created the object.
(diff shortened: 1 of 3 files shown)


lib/deimos/kafka_source.rb
```diff
@@ -88,8 +88,9 @@ def import_without_validations_or_callbacks(column_names,
                                                   array_of_attributes,
                                                   options={})
         results = super
-        return unless self.kafka_config[:import]
-        return if array_of_attributes.empty?
+        if !self.kafka_config[:import] || array_of_attributes.empty?
+          return ActiveRecord::Import::Result.new([], 0, [], [])
+        end
```
dorner, line 93: Is there a reason you can't just return `results` here?

dorner (changes requested):

lib/deimos/kafka_source.rb
DeeChau: oh good point, will try that out - i forgot that we called `super` haha

DeeChau: Yeah - the `results` from `super` works with this too

dorner (approved):

PR #101, opened 2020-11-30
GUILD-556: Retry on Offset seek error

### LINKED ISSUE #5: Inline consumer should use retries when seeking
Currently, InlineConsumer will try to seek to the necessary offset, and if it can't (e.g. LeaderNotAvailable error), it'll just ignore the seek. This may cause integration tests to fail. We should replace this logic with sleep + retries (up to a max of 3), similar to what RubyKafka does in this case.

dvvrt:
# Pull Request Template

## Description

Update inline consumer to retry offset seek up to 2 times when an error is encountered

Fixes #5 

## Type of change

## How Has This Been Tested?
-  Updated rspecs 

## Checklist:


dorner (commented):
@dvvrt looks really good! Please add an entry to the CHANGELOG under UNRELEASED and it's good to merge 👍

dorner (approved):

PR #113, opened 2021-01-22
Cpfm 246

### LINKED ISSUE #112: DB Poller should order results by record class timestamps, to prevent ambiguity from joins

SeanTCasey:
Description
The DB Poller orders results by a timestamp (like updated_at, created_at, etc.), but if a join was done between multiple tables that share these timestamps, the order query becomes ambiguous, resulting in a crash. This fix sets the owner column for the timestamp to prevent this.

Fixes #112 (issue)

Type of change

[ X ] Bug fix (non-breaking change which fixes an issue)
How Has This Been Tested?
- Existing tests pass successfully. This is a bit of an edge case with my own project, so probably doesn't warrant a full test around it on Deimos. It's just making the ordering a bit more explicit. The service that uses this logic, Content Sieve, has tests that pass with this fix.

[ X ] My code follows the style guidelines of this project
[ X ] I have performed a self-review of my own code
[ X ]I have added a line in the CHANGELOG describing this change, under the UNRELEASED heading
[ X ] My changes generate no new warnings
[ X ] New and existing unit tests pass locally with my changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## UNRELEASED
-
+- Prevents DB Poller from reconnecting to DB if there is an open transaction
 - Replaces `before` by `prepend_before` for more consistent test setups.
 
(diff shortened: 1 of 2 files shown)


dorner:
@SeanTCasey looks good but:
* lint is failing
* needs rebase
* can we add a spec where we do this via join?

PR #116, opened 2021-04-16
[GUILD 616] - Generator for dumping data from kafka to DB

hemant3402:
# Pull Request Template

## Description

1) Creates a consumer file
2) Updates the configuration to add the consumer
3) Creates a database migration to create the model
4) Creates an ActiveRecord model

## Type of change

## How Has This Been Tested?

Added a generator rspec comparing the generated files to the expected output

- [x] Rspec Test

## Checklist:


lib/generators/deimos/kafka_consumer_generator.rb
```diff
@@ -0,0 +1,60 @@
...
+
+        # @return [String]
+        def namespace
+          last_dot = self.full_schema.rindex('.')
+          self.full_schema[0...last_dot]
+        end
+
+        # @return [String]
+        def consumer_block
+          "consumer do\n  topic '#{topic_name}'\n  class_name '#{table_class}Consumer'\n  namespace '#{namespace}'\n  schema '#{schema}'\n  group_id 'PLEASE FILL THIS'\nend\n"
```
hemant3402, line 46: Wondering if adding another cli option to add group ID for the consumer. Not sure if adding too many options will confuse the users

README.md
```diff
@@ -661,6 +661,31 @@ Example:
...
+database table based on its fields, updates the specified kafka configuration file,
+and creates a consumer class.
+
+Before running this migration, you must first copy the schema into your repo
+in the correct path (in the example above, you would need to have a file
+`{SCHEMA_ROOT}/com/my-namespace/MySchema.avsc`).
+
+To generate a model and migration, run the following:
+
+    rails g deimos:kafka_consumer TABLE_NAME FULL_SCHEMA_NAME TOPIC_NAME KAFKA_CONFIG_FILE_PATH
```
dorner, line 676: For that many arguments I don't think positional arguments is going to be usable. We should use optional arguments, and provide sane defaults so people don't have to specify them all the time. Preferably something like:

`rails g deimos:kafka_consumer TOPIC_NAME FULL_SCHEMA_NAME`

where we could infer the table name by just taking the name of the schema (e.g. if the schema type is called "FlyerItem" we could create a table "flyer_items"). I think we could even figure out which initializer file does the configuration by searching the files in `config/initializers` for `Deimos.configure`. (We can override this in FlippRubyKafka to work for our version as well.)

lib/generators/deimos/kafka_consumer/templates/consumer.rb.tt
```diff
@@ -0,0 +1,8 @@
+class <%= "#{table_class}Consumer" %> < Deimos::Consumer
```
dorner, line 1: Wouldn't this be an ActiveRecordConsumer?

lib/generators/deimos/kafka_consumer_generator.rb
dorner: We should be able to generate the group ID based on the application name and the topic name.

dorner (changes requested):

hemant3402:
> @hemant3402 thanks for all the hard work with this! Unfortunately when I first created this ticket I'd completely forgotten about the existing functionality that created the table. Looking over this, the added feature of creating the consumer class and config piece are small enough (and require enough extra configuration) that I don't think it's needed.
> 
> I'm going to replace this with simply adding some text to print out after the current generator finishes suggesting how the user can do this themselves.
> 
> I'm sorry for the work that'll end up not being used, but I hope you were able to learn something while doing it!

@dorner Yep absolutely no worries, definitely learnt a lot about rails generators and THOR CLI library while doing this

PR #120, opened 2021-05-14
GUILD-619: Test helpers don't automatically reset to test config

angad-singh:
## Description

- Deimos is not reset back to the default test config before each test now.
- A function has been added which can be called by clients in tests to manually reset Deimos back to test config.
- The same function can also be used to override settings while keeping defaults in place.

Fixes https://flippit.atlassian.net/browse/GUILD-619

## Type of change

## How Has This Been Tested?

Tested by calling this function in existing specs locally.

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## UNRELEASED
+### Roadmap :car:
+
+- TestHelper does not automatically reset Deimos config before each test. #120
 
 ## 1.10.0 - 2021-03-22
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## UNRELEASED
 ### Roadmap :car:
 
-- TestHelper does not automatically reset Deimos config before each test. [#120](https://github.com/flipp-oss/deimos/pull/120)
+- TestHelper does not automatically reset Deimos config before each test. [#120](https://github.com/flipp-oss/deimos/pull/120).
+  **Please note that this is a breaking change**
 
 ## 1.10.0 - 2021-03-22
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## UNRELEASED
 ### Roadmap :car:
 
-- TestHelper does not automatically reset Deimos config before each test. #120
+- TestHelper does not automatically reset Deimos config before each test. [#120](https://github.com/flipp-oss/deimos/pull/120)
 
 ## 1.10.0 - 2021-03-22
 
--- a/README.md
+++ b/README.md
@@ -29,6 +29,7 @@ Built on Phobos and hence Ruby-Kafka.
    * [Running Consumers](#running-consumers)
    * [Metrics](#metrics)
    * [Testing](#testing)
+        * [Default Deimos config for testing](#default-deimos-config-for-testing)
         * [Integration Test Helpers](#integration-test-helpers)
    * [Utilities](#utilities)
    * [Contributing](#contributing) 
--- a/README.md
+++ b/README.md
@@ -950,7 +950,7 @@ Deimos::TestHelpers.schemas_compatible?(schema1, schema2)
 ### Test Helpers
 
 There are helper methods available to configure Deimos for different types of testing scenarios. 
-Currently there is a helper defined for unit tests and for testing Kafka releated code. You can use it as follows:
+Currently there are helpers defined for unit tests and for testing Kafka related code. You can use it as follows:
 
 ```ruby
 # The following can be added to a rpsec file so that each unit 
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -935,6 +935,18 @@ expect(message).to eq({
...
+
+# You can reset Deimos to default test config by calling this method.
+configure_deimos
+
+# The same method can be used to override settings while still using
+# deafults for other fields.
+configure_deimos(consumers: { reraise_errors: false },
+                 producers: { topic_prefix: nil },
+                 db_producer: { compact_topics: %w(my-topic my-topic2) },
+                 logger: Rails.logger)
```
dorner, line 949: Nice, I like this a lot! I think there might be some other changes to the README that are necessary - e.g. we should be calling out things like the test producer and how to interact with it rather than assuming that when you include TestHelpers you get it for free.

dorner (changes requested):

README.md
angad-singh: @dorner I tried my best to add more info but there might be some that I missed and/or added incorrect information. Please let me know!

README.md
```diff
@@ -935,6 +936,18 @@ expect(message).to eq({
   topic: 'my-topic',
   key: 'my-id'
 })
+
+## Configuring Deimos to test settings
```
dorner, line 940: So this shouldn't be in the code block with everything else because it likely won't be in the test itself. Also, the previous examples in this block will *only* work if the test producers are being used, so we should call that out and put this explanation before that block. 

You can give an example of a `spec_helper` that does it:

```ruby
# spec_helper.rb
RSpec.before(:each) do
  configure_deimos
end
```

Actually, now that I write this, it would be even better if we can allow this method to take a block and reset the config back to what it was. Then you could use it like this:

```ruby
RSpec.around(:each) do |ex|
  configure_deimos({...}) do
    ex.run
  end
```

...and we would be able to ensure that config didn't leak past examples, which is something that's bitten me multiple times. You can deep-dup the config and restore it after the block is run.

dorner (changes requested):

lib/deimos/test_helpers.rb
```diff
@@ -18,22 +18,31 @@ class << self
       def sent_messages
         Deimos::Backends::Test.sent_messages
       end
+
+      # Set the config to the right settings for a unit test
+      def unit_test!
```
dorner, line 23: Can we document these as well?

dorner (changes requested):

README.md
```diff
@@ -946,6 +947,32 @@ require 'deimos/test_helpers'
...
+around(:each) do |example|
+  Deimos::TestHelpers.unit_test!
+  example.run
+  Deimos.config.reset
+end
+
+
+# Similarly you can use the Kafka test helper
+around(:each) do |example|
+  Deimos::TestHelpers.kafka_test
```
dorner, line 967: missing !

README.md
```diff
@@ -946,6 +947,32 @@ require 'deimos/test_helpers'
...
+  example.run
+  Deimos.config.reset
+end
+```
+
+With the help of these helper methods, rspec examples can be written without having to tinker with Deimos settings.
+This also prevents Deimos setting changes from leaking in to other examples.
```
dorner, line 974: Maybe add an example for how people can configure Deimos manually as well without using these new methods.

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## UNRELEASED
+### Roadmap :car:
+
+- TestHelper does not automatically reset Deimos config before each test. [#120](https://github.com/flipp-oss/deimos/pull/120)
```
dorner, line 11: Should call out in bold that this is a breaking change.

lib/deimos/test_helpers.rb
```diff
@@ -18,22 +18,31 @@ class << self
...
+          deimos_config.producers.backend = :test
+        end
+      end
+
+      # Set the config to the right settings for a kafka test
+      def kafka_test!
+        Deimos.configure do |deimos_config|
+          deimos_config.producers.backend = :kafka
+          deimos_config.schema.backend = :avro_validation
+        end
```
dorner, line 38: Maybe we should have one more that uses the full schema registry instead of avro_validation?

dorner (changes requested):

README.md
```diff
@@ -946,6 +947,55 @@ require 'deimos/test_helpers'
...
+# Similarly you can use the Kafka test helper
+around(:each) do |example|
+  Deimos::TestHelpers.kafka_test!
+  example.run
+  Deimos.config.reset
+end
+
+# Kakfa test helper using schema registry
+around(:each) do |example|
+  Deimos::TestHelpers.kafka_schema_registry_test!
```
dorner, line 973: Maybe call this `full_integration_test`?

README.md
```diff
@@ -946,6 +947,55 @@ require 'deimos/test_helpers'
...
+```ruby
+    it 'should not fail this random test' do
+      
+      Deimos.configure do |config|
+        config.consumers.fatal_error = proc { true }
+        config.consumers.reraise_errors = false
+      end
+      ...
+      expect(some_object).to be_truthy
+      ...
```
dorner, line 992: Add a `Deimos.config.reset` here too :)

dorner (approved):

PR #122, opened 2021-06-16
GUILD-522: Add ability to generate schema model classes from schemas

DeeChau:
## Description

New Feature - Schema Class Generation

Offer a way to generate classes from Avro schemas. These classes are documented with YARD to aid in IDE auto-complete, and will help to move errors closer to the code. These classes can be used in your application through the Consumer and Producer.

## Type of change

## How Has This Been Tested?

- [x] Created new and updated existing unit tests in Deimos
- [x] Set up in applications currently using Deimos - verified that existing functionality does not break

## Checklist:

- [] I have commented my code, particularly in hard-to-understand areas

## Suggested order of files to look at

* README.md, CONFIGURATION.md
* Generate Schema classes from Avro Schemas
  * schema_record.rb, schema_enum.rb
  * schema_class_generator.rb, schema_class_mixin.rb
  * schema_record.rb.tt, schema_enum.rb.tt
  * spec/schema_classes/ - Examples of classes generated from schemas
  * spec/generators/schema_class_generator_spec.rb
  * spec/generators/schema_class/
* Use Schema Classes in your consumer/producer interfaces
  * configuration.rb
  * consumer.rb, message_consumption.rb, batch_consumption.rb
  * active_record_consume/batch_consumption.rb
  * spec/consumer_spec.rb, spec/batch_consumer_spec.rb
  * spec/active_record_consumer_spec.rb, spec/active_record_batch_consumer_spec.rb
  * producer.rb, active_record_producer.rb
  * spec/active_record_producer_spec.rb, spec/producer_spec.rb
* Miscellaneous
  * guardfile generation


lib/generators/deimos/schema_class/templates/schema_record.rb.tt
```diff
@@ -0,0 +1,68 @@
...
+
+    # @override
+    def self.initialize_from_payload(payload)
+      return unless payload.present?
+
+      args = {}
+      payload.each do |key, value|
+<%- unless @special_field_initialization.present? -%>
+        args[key.to_sym] = value
+<%- else -%>
```
dorner, line 34: Can you do `if/else` instead of `unless/else`?

lib/generators/deimos/schema_class/templates/schema_record.rb.tt
```diff
@@ -0,0 +1,68 @@
...
+      super()
+      args = kwargs.with_indifferent_access
+<%- fields.each do |field| -%>
+      <%= "@#{field.name} = args[:#{field.name}]" %>
+<% end -%>
+      @payload_key = args[:payload_key]
+    end
+
+    # @override
+    def self.initialize_from_payload(payload)
```
dorner, line 27: I don't really understand why this is a different method from `initialize`? A hash and a set of keyword args are essentially the same thing?

lib/generators/deimos/schema_class/templates/schema_record.rb.tt
```diff
@@ -0,0 +1,68 @@
...
+                           when <%= fields.join(', ') %>
+                             <%= initialize_string %>
+<% end -%>
+                           else
+                             value
+                           end
+<%- end -%>
+      end
+      self.new(**args)
+    end
```
dorner, line 46: I feel like this might be simpler if you just generate a `def #{field}` and `def #{field}=` method rather than relying on `attr_accessor`. This works but it seems a bit odd.

It would also make it easier to set default values! 😁

lib/generators/deimos/schema_class_generator.rb
```diff
@@ -0,0 +1,231 @@
...
+        # @return [String] A string representation of the Type of this SchemaField
+        def deimos_field_type(schema_field)
+          _field_type(schema_field.type)
+        end
+
+        # @param enum [Avro::Schema::EnumSchema] a field of type 'enum'.
+        # @return [Array<String>] of symbols valid for the enum.
+        def enum_symbols(enum)
+          enum.symbols
+        end
```
dorner, line 42: Seems a bit redundant? :)

lib/generators/deimos/schema_class_generator.rb
```diff
@@ -0,0 +1,231 @@
...
+        # @return [String] the possible return values for this Enum type
+        def enum_return_values(enum)
+          "'#{enum.symbols.join("', '")}'"
+        end
+
+        # Generate a Schema Model Class and all of its Nested Records from an Avro Schema
+        # @param schema_name [String] the name of the Avro Schema in Dot Syntax
+        def generate_classes_from_schema(schema_name)
+          _schema_base(schema_name).load_schema!
+          _schema_base.schema_store.schemas.each_value do |schema|
```
dorner, line 54: Wondering if it would make more sense to read the Deimos config rather than the file system for this.

lib/generators/deimos/schema_class_generator.rb
```diff
@@ -0,0 +1,231 @@
...
+
+      # @param avro_schema [Avro::Schema::NamedSchema]
+      # @return [String]
+      def _field_initialize_formatting(avro_schema)
+        field_type = _field_type(avro_schema)
+        case avro_schema.type_sym
+        when :record
+          "#{field_type}.initialize_from_payload"
+        when :enum
+          "#{field_type}.new"
```
dorner, line 224: same question above - either we don't need `initialize_from_payload` and we can just use `new`, or we could just define an `initialize_from_payload` method on the enum and dispense with this method.

lib/generators/deimos/schema_class_generator.rb
```diff
@@ -0,0 +1,231 @@
...
+                                else
+                                  "#{initialize_method}(value)"
+                                end
+
+            result[initialize_string] << ":#{field.name}"
+          end
+          result
+        end
+
+        # Format a given field into it's appropriate to_h representation.
```
dorner, line 91: "its" 😁

lib/generators/deimos/schema_class_generator.rb
```diff
@@ -0,0 +1,231 @@
...
+            filename = "#{Deimos.config.schema.generated_class_path}/#{namespace_path}/#{file_prefix}.rb"
+            template(schema_template, filename, force: true)
+          end
+          _clear_schema_base!
+        end
+
+        # Retrieve any special formatting needed for this current schema's fields
+        # @return [Hash<String, Array[Symbol]>]
+        def special_field_initialization
+          result = Hash.new { |h, k| h[k] = [] }
```
dorner, line 69: CHECK THIS OUT - you can pass the default value into the hash constructor!
`result - Hash.new([])`

lib/generators/deimos/schema_class_generator.rb
```diff
@@ -0,0 +1,231 @@
...
+          fields.each do |field|
+            field_base_type = field.type.type_sym # Record, Union, Enum, Array or Map?
```
dorner, line 71: Yeah there's a lot of Avro-specific stuff here. Let's see if we can move some of it into the backend base. Hopefully it can be as simple as "here's a schema, here's a field, is the Ruby object an array, an object, an enum or a scalar value?"

lib/generators/deimos/schema_class_generator.rb
```diff
@@ -0,0 +1,231 @@
...
+          gsub('/', '.')
+
+        "#{namespace_dir}.#{schema_name}"
+      end
+
+      # Converts Avro::Schema::NamedSchema's to String form for generated YARD docs.
+      # Recursively handles the typing for Arrays, Maps and Unions.
+      # @param avro_schema [Avro::Schema::NamedSchema]
+      # @return [String] A string representation of the Type of this SchemaField
+      def _field_type(avro_schema)
```
dorner, line 174: E.g. this method could go into avro_base.

lib/generators/deimos/schema_class_guardfile_generator.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
```
dorner, line 1: I'm really not sure we need the Guardfile. Schemas change really rarely and if they change and people are using the record objects, it's likely they'll notice it pretty quickly.

dorner (changes requested):

docs/CONFIGURATION.md
```diff
@@ -81,6 +82,7 @@ namespace|nil|Namespace of the schema to use when finding it locally.
 key_config|nil|Configuration hash for message keys. See [Kafka Message Keys](../README.md#installation)
 disabled|false|Set to true to skip starting an actual listener for this consumer on startup.
 group_id|nil|ID of the consumer group.
+use_schema_class|nil|Set to false to disable using the consumers schema classes. See [Generated Schema Classes](../README.md#generated-schema-classes)
```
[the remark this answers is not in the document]
DeeChau: No particular reason, it's been a while since this feature was introduced but it can turn it on too if you don't have the schema level one on

lib/deimos/producer.rb
```diff
@@ -113,7 +116,10 @@ def publish_list(payloads, sync: nil, force_send: false, topic: self.topic)
           topic: topic,
           payloads: payloads
         ) do
-          messages = Array(payloads).map { |p| Deimos::Message.new(p, self) }
+          messages = Array(payloads).map do |payload|
+            p = payload.is_a?(SchemaRecord) ? payload.to_h : payload
```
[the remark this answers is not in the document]
DeeChau: i think nil was the reason - some rspecs broke cause of it, will comment it

lib/deimos/schema_backends/avro_base.rb
```diff
@@ -77,6 +77,11 @@ def validate(payload, schema:)
                                         fail_on_extra_fields: true)
       end
 
+      # :nodoc:
+      def load_schema!
```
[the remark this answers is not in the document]
DeeChau: I'm assuming that you mean `schema_backends/base.rb`.

I actually don't remember why i used it.. I'll look at if this is actually needed, remove if it's not and update the name if it is useful!

lib/deimos/test_helpers.rb
```diff
@@ -51,6 +51,16 @@ def kafka_test!
...
 
+        config.before(:suite) do
+          Deimos.configure do |d_config|
+            d_config.logger = Logger.new($stdout)
+            d_config.consumers.reraise_errors = true
+            d_config.kafka.seed_brokers ||= ['test_broker']
+            d_config.schema.backend = Deimos.schema_backend_class.mock_backend
+            d_config.producers.backend = :test
+          end
+        end
```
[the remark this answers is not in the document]
DeeChau: sure was, how did THIS get here 🤔 ?? removed it and nothing broke

lib/deimos/utils/schema_class_mixin.rb
```diff
@@ -0,0 +1,52 @@
...
+
+        klass.initialize_from_payload(payload)
+      end
+
+      # @param config [FigTree::ConfigStruct] Producer or Consumer config
+      # @return [Boolean]
+      def use_schema_class?(config)
+        use_schema_class = config[:use_schema_class]
+        use_schema_class.present? ? use_schema_class : Deimos.config.schema.use_schema_class
+      end
```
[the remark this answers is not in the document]
DeeChau: Hrmm - yeah can move it to the `avro_base` class

lib/generators/deimos/schema_class/templates/schema_record.rb.tt
```diff
@@ -0,0 +1,68 @@
...
+    # @return [<%= deimos_field_type(field) %>]
+    attr_accessor :<%= field.name %>
+<% end -%>
+<%- unless @current_schema.name =~ /\-key$/ -%>
+    # @return [Object] An optional payload key
+    attr_accessor :payload_key
+<% end -%>
+
+    # @override
+    def initialize(**kwargs)
```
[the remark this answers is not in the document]
DeeChau: Matthew actually started at that approach first - i can look into reverting it to that!

lib/generators/deimos/schema_class/templates/schema_record.rb.tt
DeeChau: Yeap

lib/generators/deimos/schema_class/templates/schema_record.rb.tt
DeeChau: The only difference is that it's handling the hashes recursively and initializing sub-records/enums/maps/arrays whereas `initialize` would accept the already formatted values. I think i need to write an example of how this would be used

lib/generators/deimos/schema_class_generator.rb
DeeChau: lol thats silly, i _think_ matthew left it over, will clean up this and other ones haha

lib/generators/deimos/schema_class_generator.rb
DeeChau: I was actually thinking of that recently, i could possibly use the `Deimos.config.consumer_objects` and `Deimos.config.producer_objects` to achieve this

My quick investigating showed that this might be a bit of work to get a working solution though, but would be promising as the requirements for those objects are to set the schema and other things

lib/generators/deimos/schema_class_generator.rb
DeeChau: yeah, that would be useful

lib/generators/deimos/schema_class_generator.rb
DeeChau: WHAAAAAAAAAAT?!?!?!? NBO WA\Y

lib/generators/deimos/schema_class_generator.rb
DeeChau: noice - i like this method! lets promote it

lib/generators/deimos/schema_class_guardfile_generator.rb
DeeChau: good point, can look into removing it :)

README.md
```diff
@@ -795,6 +796,102 @@ which can be useful if you want to figure out if you're inside the task
...
+class MyActiveRecordConsumer < Deimos::ActiveRecordConsumer
+  record_class Widget
+  # The following methods will treat payload as an instance of Deimos::SchemaRecord rather than a hash.
+  def fetch_record(klass, payload, key) ; super end
+  def assign_key(record, payload, key) ; super end
+  def record_attributes(payload, key)
+    super.merge(:some_field => "some_value-#{payload.test_id}")
+  end
+  def record_key(payload) ; super end
+  def process_message?(payload) ; super end
```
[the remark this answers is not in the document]
DeeChau: Adjusted docs for these methods too :)

lib/deimos/producer.rb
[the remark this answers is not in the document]
DeeChau: Works now, looks like this wasn't needed. I think i actually fixed the behaviour, but didn't get to fix this part

lib/deimos/schema_backends/avro_base.rb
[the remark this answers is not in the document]
DeeChau: Ah i remember now - it's cause `avro_schema` was a private method and i wanted to get around it. WIll look into refactoring this with the below comments to add it to the `avro_base` class :)

lib/generators/deimos/schema_class/templates/schema_enum.rb.tt
```diff
@@ -0,0 +1,31 @@
...
+# autogenerated by Deimos, do not modify
+module Deimos
+  # :nodoc:
+  class <%= schema_classname(@current_schema) %> < SchemaEnum
+    # @return [<%= enum_return_values(@current_schema) %>]
+    attr_accessor :<%= @current_schema.name.underscore %>
+
+    # :nodoc:
+    def initialize(<%= @current_schema.name.underscore %>)
+      super()
```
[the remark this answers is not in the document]
DeeChau: Look like i don't really need super at all - will add back if needed

lib/generators/deimos/schema_class/templates/schema_record.rb.tt
DeeChau: hahaha good idea

lib/generators/deimos/schema_class_generator.rb
DeeChau: didnt work - will look into later >:(

lib/generators/deimos/schema_class/templates/schema_enum.rb.tt
```diff
@@ -0,0 +1,31 @@
...
+      @<%= @current_schema.name.underscore %> = <%= @current_schema.name.underscore %>
+    end
+
+    # @override
+    def symbols
+      %w(<%= enum_symbols(@current_schema).join(' ') %>)
+    end
+
+    # @override
+    def to_json(*_args)
```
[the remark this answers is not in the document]
DeeChau: Doesn't look like we actually _need_ to override either as_json or to_json for schema_enum, there's a later revision that fixes this

lib/generators/deimos/schema_class/templates/schema_record.rb.tt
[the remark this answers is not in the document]
DeeChau: Done with wrapping!

lib/generators/deimos/schema_class_generator.rb
DeeChau: Hrmm, this may be challenging to set up with rspecs - may need to mock it out quite a bit for that.

lib/generators/deimos/schema_class/templates/schema_record.rb.tt
DeeChau: Looking into handling recursive initialization via. the `initialize` method

lib/generators/deimos/schema_class/templates/schema_record.rb.tt
DeeChau: Using `attr_readers` and created custom `def #{field}=` methods for fields that need this special handling!

using `attr_accessors` for everything else!

lib/generators/deimos/schema_class_generator.rb
DeeChau: Removed this function ;)

lib/generators/deimos/schema_class_generator.rb
DeeChau: Gonna reserve this for last, got most of the other requested changes in!

DeeChau: Looking into this for the next 15% time. Have most/all of the other comments addressed!

lib/deimos/utils/schema_class_mixin.rb
```diff
@@ -0,0 +1,32 @@
...
+
+      # @param config [Hash] Producer or Consumer config
+      # @return [Boolean]
+      def use_schema_classes?(config)
+        use_schema_classes = config[:use_schema_classes]
+        use_schema_classes.present? ? use_schema_classes : Deimos.config.schema.use_schema_classes
+      end
+    end
+  end
+end
```
dorner, line 32: Is there a reason this is a mixin instead of a module that's used on its own (using `class << self`)? None of these methods use the current class or any state, and it's a bit cleaner to have it used as a separate module.

lib/generators/deimos/schema_class/templates/schema_enum.rb.tt
```diff
@@ -0,0 +1,26 @@
...
+module Deimos
+  # :nodoc:
+  class <%= Deimos::SchemaBackends::AvroBase.schema_classname(@current_schema) %> < SchemaClass::Enum
+    # @return ['<%= @current_schema.symbols.join("', '") %>']
+    attr_accessor :<%= @current_schema.name.underscore %>
+
+    # :nodoc:
+    def initialize(<%= @current_schema.name.underscore %>)
+      super()
+      self.<%= @current_schema.name.underscore %> = <%= @current_schema.name.underscore %>
```
dorner, line 13: see above conversation, what happened there?

lib/generators/deimos/schema_class/templates/schema_record.rb.tt
```diff
@@ -0,0 +1,83 @@
...
+<%- else -%>
+      @<%= method_definition[:field].name %> = <%= method_definition[:field_initialization] %>
+<% end -%>
+    end
+
+<% end -%>
+
+    # @override
+    <%= @initialization_definition %>
+      super()
```
dorner, line 48: remove `()`?

lib/deimos/schema_backends/base.rb
```diff
@@ -3,15 +3,16 @@
...
   # Represents a field in the schema.
   class SchemaField
-    attr_accessor :name, :type, :enum_values
+    attr_accessor :name, :type, :enum_values, :default
 
     # @param name [String]
     # @param type [Object]
     # @param enum_values [Array<String>]
-    def initialize(name, type, enum_values=[])
+    def initialize(name, type, enum_values=[], default=:no_default)
```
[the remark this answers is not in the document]
dorner: Actually, is there any practical difference between `nil` and `no_default`?

spec/schema_classes/a_record.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+# This file is autogenerated by Deimos, Do NOT modify
+module Deimos
```
dorner, line 4: Just realized something... are these being generated as separate files? We might run into issues if there are multiple Avro schemas that have records defined in them and those records' names conflict. Probably better to generate them as inner classes. :(

dorner (changes requested):

lib/generators/deimos/schema_class/templates/schema_record.rb.tt
DeeChau: doesn't work without `()` :( I think it's because we're using Keyword args and technically there are **no** args

```ruby
     Failure/Error: super

     ArgumentError:
       wrong number of arguments (given 1, expected 0)
```

lib/deimos/schema_backends/avro_base.rb
```diff
@@ -135,6 +199,20 @@ def _field_name_from_schema(value_schema)
       def _key_schema_name(schema)
         "#{schema}_key"
       end
+
+      # @return [Boolean] If the schema is being used in a Consumer
+      def consumer_key_schema?
+        @consumer_key_schemas ||= Deimos.config.consumer_objects.map { |c| c.try(:key_schema) }.
+          uniq.compact
+        @consumer_key_schemas.include?(avro_schema.name)
+      end
```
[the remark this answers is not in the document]
DeeChau: will take a stab! Should be possible

lib/deimos/schema_backends/base.rb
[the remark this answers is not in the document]
DeeChau: so, by default (haha)

the Avro class's `Avro::Field` stores `:no_default` in the `default` accessor

image

So to "translate" that over on our side, the "default" for default is `:no_default`.

I can however, do what you mentioned to make this more explicit and clear that this is a sentinal value - like making it  a method or a constant `NO_DEFAULT`

lib/deimos/schema_class/base.rb
```diff
@@ -0,0 +1,62 @@
...
+      # Converts the object to a string that represents a JSON object
+      # @return [String] a JSON string
+      def to_json(*_args)
+        to_h.to_json
+      end
+
+      # Converts the object to a hash which can be used for debugging.
+      # @return [Hash] a hash representation of the payload
+      def as_json(_opts={})
+        JSON.parse(to_json)
```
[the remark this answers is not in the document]
DeeChau: yeah that works now - i think it's cause i rewrote `to_h` and the implemented versions to flatten any instances of classes :D

lib/deimos/schema_class/base.rb
```diff
@@ -0,0 +1,62 @@
...
+      end
+
+      # :nodoc:
+      def ==(other)
+        comparison = other
+        if other.class == self.class
+          comparison = other.state
+        end
+
+        comparison == self.state
```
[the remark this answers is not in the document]
DeeChau: oo thats silly of me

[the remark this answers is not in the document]
DeeChau: changing to to_h :D

lib/deimos/schema_backends/base.rb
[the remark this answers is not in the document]
dorner: Nah, if it's good enough for Avro it's good enough for me. :p

lib/generators/deimos/schema_class/templates/schema_enum.rb.tt
DeeChau: Removing these causes breakages !! - Discussion is below!

lib/deimos/utils/schema_class_mixin.rb
DeeChau: will look to extract into a module

DeeChau: Am not sure what to do here, might need a bit of help tomorrow to go over making this cleaner

lib/deimos/schema_backends/avro_base.rb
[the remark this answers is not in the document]
DeeChau: With the recent update to use Deimos Consumer objects - this is actually not used anymore!

I'm thinking of removing it for the sake of eliminating unused code and keeping this class cleaner

lib/generators/deimos/schema_class/templates/schema_record.rb.tt
DeeChau: Found a related article - https://www.rubyguides.com/2018/09/ruby-super-keyword/

> But what if the parent method doesn't take the same number of arguments?
In that case you can use:
`super()` for no arguments
`super(arg1, arg2, ...)` to choose what arguments you want to pass

Makes me wonder if we _should_ include `super()` here at all - the base classes (record, enum and base) don't do any special initialization _yet_

lib/deimos/schema_backends/base.rb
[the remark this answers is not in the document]
DeeChau: sounds good to me!!

lib/generators/deimos/schema_class/templates/schema_record.rb.tt
dorner: If `super` fails then we're defining these classes wrong somehow - either the base or the child classes have the method defined differently.

spec/schema_classes/a_record.rb
DeeChau: i _think_ i can do something similar and create a template for inner classes (Using something like `insert_into_file` or just plain ERB template loading from a helper method :p) - will probably shove all the inner classes _above_ the original logic

dorner: Yep, something like that.

DeeChau: Here's an example of what i've got so far :)

```ruby
# frozen_string_literal: true

# This file is autogenerated by Deimos, Do NOT modify
module Deimos
  # Primary Schema Class
  # Autogenerated Schema for Record at com.my-namespace.MyNestedSchema
  class MyNestedSchema < SchemaClass::Record
    # Attribute Readers
    # @return [Deimos::MyNestedRecord]
    attr_reader :some_nested_record
    # @return [nil, Deimos::MyNestedRecord]
    attr_reader :some_optional_record
    # Attribute Accessors
    # @param value [String]
    attr_accessor :test_id
    # @param value [Float]
    attr_accessor :test_float
    # @param values [Array<String>]
    attr_accessor :test_array
    # @return [Object] An optional payload key
    attr_accessor :payload_key
    # Attribute Writers
    # @param value [Deimos::MyNestedRecord]
    def some_nested_record=(value)
      @some_nested_record = Deimos::MyNestedRecord.initialize_from_value(value)
    end

    # @param value [nil, Deimos::MyNestedRecord]
    def some_optional_record=(value)
      @some_optional_record = Deimos::MyNestedRecord.initialize_from_value(value)
    end


    # @override
    def initialize(test_id: nil,
                   test_float: nil,
                   test_array: nil,
                   some_nested_record: nil,
                   some_optional_record: nil,
                   payload_key: nil)
      super()
      self.test_id = test_id
      self.test_float = test_float
      self.test_array = test_array
      self.some_nested_record = some_nested_record
      self.some_optional_record = some_optional_record
      self.payload_key = payload_key
    end

    # @override
    def schema
      'MyNestedSchema'
    end

    # @override
    def namespace
      'com.my-namespace'
    end

    # @override
    def to_h
      payload = {
        'test_id' => @test_id,
        'test_float' => @test_float,
        'test_array' => @test_array,
        'some_nested_record' => @some_nested_record&.to_h,
        'some_optional_record' => @some_optional_record&.to_h
      }
      @payload_key.present? ? payload.merge('payload_key' => @payload_key) : payload
    end

    # Secondary Schema Classes
      # Autogenerated Schema for Record at com.my-namespace.MyNestedRecord
    class MyNestedRecord < SchemaClass::Record
      # Attribute Accessors
      # @param value [Integer]
      attr_accessor :some_int
      # @param value [Float]
      attr_accessor :some_float
      # @param value [String]
      attr_accessor :some_string
      # @param value [nil, Integer]
      attr_accessor :some_optional_int
      # @return [Object] An optional payload key
      attr_accessor :payload_key

      # @override
      def initialize(some_int: nil,
                     some_float: nil,
                     some_string: nil,
                     some_optional_int: nil,
                     payload_key: nil)
        super()
        self.some_int = some_int
        self.some_float = some_float
        self.some_string = some_string
        self.some_optional_int = some_optional_int
        self.payload_key = payload_key
      end

      # @override
      def schema
        'MyNestedRecord'
      end

      # @override
      def namespace
        'com.my-namespace'
      end

      # @override
      def to_h
        payload = {
          'some_int' => @some_int,
          'some_float' => @some_float,
          'some_string' => @some_string,
          'some_optional_int' => @some_optional_int
        }
        @payload_key.present? ? payload.merge('payload_key' => @payload_key) : payload
      end
    end
  
  end
end
```

dorner: Good start! I don't think I would nest the generated modules under `Deimos` because these are part of the app, not Deimos itself. Also, I think you'd need to put the nested definitions above the main one, because otherwise you'll get a crash when it tries to use them :)

DeeChau: ooooo really - can move above, would it look something like this?

```ruby
# This file is autogenerated by Deimos, Do NOT modify

# Secondary Schema Classes
# Autogenerated Schema for Record at com.my-namespace.MyNestedRecord
class MyNestedRecord < Deimos::SchemaClass::Record
  ...
end

# :nodoc:
module Deimos
  # Primary Schema Class
  # Autogenerated Schema for Record at com.my-namespace.MyNestedSchema
  class MyNestedSchema < SchemaClass::Record
    ...
  end
end
```

lib/generators/deimos/schema_class/templates/schema_record.rb.tt
DeeChau: hrm - will try something out :) - gonna define empty `initialize`'s for everything

DeeChau: so this only works if i have the following in my base class
```ruby
      # :nodoc:
      def initialize(_)
      end
```

i need to manually define the initialize method with a single argument `_` before i can change `super()` to `super`

spec/schema_classes/a_record.rb
DeeChau: Pushed le change for this and some of the resolved ones above!

lib/generators/deimos/schema_class_generator.rb
```diff
@@ -42,28 +43,35 @@ def deimos_field_type(schema_field)
         # @param key_schema_name [String]
         def generate_classes(schema_name, namespace, key_schema_name)
           schema_base = Deimos::SchemaBackends::AvroBase.new(schema: schema_name, namespace: namespace)
-          generate_classes_from_schema_base(schema_base)
+          schema_base.load_schema
+          if key_schema_name.present?
+            key_schema_base = Deimos::SchemaBackends::AvroBase.new(schema: key_schema_name, namespace: namespace)
```
dorner, line 48: Is there a reason we have to hard code this to Avro? We should be able to use `Deimos.schema_backend`

lib/generators/deimos/schema_class_generator.rb
```diff
@@ -197,6 +230,106 @@ def _field_type(avro_schema)
       def _schema_base_class(avro_schema)
         Deimos::SchemaBackends::AvroBase.schema_base_class(avro_schema)
       end
+
+      # An ERB template for schema record classes
+      # @return [String]
+      def _record_class_template
```
dorner, line 236: erm please put this in a file and read it in rather than pasting the code in the generator?

DeeChau: you sir are smart!

lib/generators/deimos/schema_class_generator.rb
```diff
@@ -197,6 +230,106 @@ def _field_type(avro_schema)
...
+      \}
+    end
+  end
+        }.strip
+      end
+
+      # An ERB template for schema enum classes
+      # @return [String]
+      def _enum_class_template
+        %{
```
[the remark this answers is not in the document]
DeeChau: yeah

lib/generators/deimos/schema_class_generator.rb
DeeChau: Yeah - i _validate_ currently to enforce Avro, can change after i update the generator rspec to include
`deimos_config.schema.backend = :avro_schema_registry` or something

DeeChau: yeap - this is doable!

lib/generators/deimos/schema_class/templates/schema_record.rb.tt
DeeChau: @dorner tagging for visibility! i can make the change mentioned above for `def initialize(_)` but am unsure if that's proper definition for a base/child class

lib/deimos/utils/schema_class_mixin.rb
DeeChau: @dorner tagging for visibility!!

lib/deimos/utils/schema_class.rb
```diff
@@ -0,0 +1,29 @@
...
+
+          klass.new(**payload.symbolize_keys)
+        end
+
+        # @param config [Hash] Producer or Consumer config
+        # @return [Boolean]
+        def use_schema_classes?(config)
+          use_schema_classes = config[:use_schema_classes]
+          use_schema_classes.present? ? use_schema_classes : Deimos.config.schema.use_schema_classes
+        end
```
dorner, line 24: Only comment here is that the words "schema class" are repeated whenever you use these methods - e.g. `SchemaClass.schema_class_instance`. I'd reword it to just `instance` and `use?`.

lib/deimos/schema_class/base.rb
```diff
@@ -0,0 +1,67 @@
...
+
+require 'json'
+
+module Deimos
+  module SchemaClass
+    # Base Class for Schema Classes generated from Avro.
+    class Base
+
+      # :nodoc:
+      def initialize(_)
```
dorner, line 11: Does it work if you replace `_` with `*args`?

dorner (changes requested):

lib/deimos/schema_class/base.rb
DeeChau: gonna try!

DeeChau: yeap sure does - gonna use `*_args`

lib/deimos/utils/schema_class.rb
DeeChau: good point! will reword

dorner (approved):

dorner:
LGTM! Awesome job Derek, I think this is one of the biggest changes we've ever made, and definitely the biggest done by anybody other than me. :D

PR #126, opened 2021-11-01
Fix for changing Deimos:: to Schema:: module namespace

DeeChau:
# Pull Request Template

## Description

Fixes issue that was released with #122 - `Deimos` namespace was used previously for generated Schema Classes. Updated to use the `Schema` namespace and updated rspecs that were previously missed.

## Type of change

## How Has This Been Tested?

Tested on applications that were not working with #122. Fixed on those apps.

## Checklist:

### DIFF
--- a/lib/deimos/schema_class/record.rb
+++ b/lib/deimos/schema_class/record.rb
@@ -7,14 +7,11 @@ module Deimos
   module SchemaClass
     # Base Class of Record Classes generated from Avro.
     class Record < Base
-      # :nodoc:
-      def ==(other)
-        comparison = other
-        if other.class == self.class
-          comparison = other.state
-        end
 
-        comparison.stringify_keys.except('payload_key') == self.state.except('payload_key')
+      # Converts the object to a hash which can be used for debugging or comparing objects.
+      # @return [Hash] a hash representation of the payload
+      def as_json(_opts={})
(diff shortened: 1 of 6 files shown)


lib/deimos/schema_class/record.rb
```diff
@@ -7,6 +7,16 @@ module Deimos
...
     # Base Class of Record Classes generated from Avro.
     class Record < Base
+      # :nodoc:
+      def ==(other)
+        comparison = other
+        if other.class == self.class
+          comparison = other.state
+        end
+
+        comparison.stringify_keys.except('payload_key') == self.state.except('payload_key')
```
dorner, line 17: Would it make more sense to just override `state` so that `payload_key` isn't output using it?

Second question... why do we have a method `state` when we could just use `as_json`? Missed this in the earlier PR.

dorner (changes requested):

lib/deimos/schema_class/record.rb
DeeChau: huh, good question :| as_json would work perfectly

DeeChau: Doing both - with `as_json`, it's not used anywhere else

dorner (approved):

Issue #153, opened 2022-04-08
Sub-schemas can conflict with each other


iMacTia:
I'm trying to understand what's the best way to manage schemas with similar/identical sub-schemas.
For example, imagine to have 2 (or more!) separate schemas, both of which use an enum like the following:

```json
{
  "doc": "The unit of the search_radius",
  "name": "search_radius_unit",
  "type": [
    "null",
    {
      "name": "radius_unit",
      "symbols": ["km", "mi"],
      "type": "enum"
    }
  ]
}
```

Now if I run `bundle exec rake deimos:generate_schema_classes`, I'll get a different class per schema, each including the sub-schema for the enum:

```ruby
# Autogenerated Schema for Enum at com.indeed.flex.workers.events.activity.radius_unit
class RadiusUnit < Deimos::SchemaClass::Enum
  # @return ['km', 'mi']
  attr_accessor :radius_unit

  # :nodoc:
  def initialize(radius_unit)
    super
    self.radius_unit = radius_unit
  end

  # @override
  def symbols
    %w(km mi)
  end

  # @override
  def to_h
    @radius_unit
  end
end
```

Since all schemas are created inside the same namespace (`Schemas`), that means the two auto-generated Enum schema classes will have the exact Ruby class path (`Schemas::RadiusUnit`). Effectively, this is causing one to reopen and override the other when Ruby loads both classes.

That is already an issue by itself (needless repetition), but it could get even worse if you consider the case where the two enums are called the same, but not exactly the same.
Imagine the second schema has a different `symbols` value, like `"symbols": ["kilometers", "miles"]`.

**Now we end up with 2 ruby classes called exactly the same, but with a different implementation of the `symbols` method.**
And since the last loaded class overrides the former(s), that means one of the two classes is now inconsistent with the corresponding schema, causing it to prevent sending messages.

cc @DeeChau

dorner:
Hmm... I feel like this is something that probably should be configurable. I can see there being value in having a top-level enum constant (e.g. when you know that the enums actually are identical). On the other hand, when they aren't identical, it might be better to generate the enum class inside the top-level class instead and therefore it would be namespaced (so you'd get `Schema1::RadiusUnit` and `Schema2::RadiusUnit`.

The problem is that it's kind of hard to configure this behavior on a per-schema basis. Probably to start with we'd want it to be either on or off.

iMacTia:
I totally appreciate this is a complex issue to solve, and I'm happy to help brainstorm about it.
[...]

PR #154, opened 2022-05-13
CHANGE: Nest or flatten sub-schemas in generated classes

dorner:
# Pull Request Template

## Description

This change is one approach to #153 . This doesn't allow for changing *namespaces* but will fix issues around nested schemas.

There were a number of problems with schema class generation:

1. The default was to generate classes in a `schema_classes` folder - but the module containing the classes is `Schemas`. This won't work with zeitwerk-powered Rails.
2. Nested schemas would live in the same *file* as the parent schema, but would still be under the main `Schemas` namespace. This not only makes the classes impossible to find by Zeitwerk, but also would clobber them if you happened to have two identically-named sub-schemas in different top-level schemas.
3. The filenames generated end in `s` when zeitwerk expects them to be singularized.

The solution is to add a new `nest_schema_classes` configuration option. When set to `true` (the default), it will nest the sub-schema classes inside their parent schema classes. This ensures that no clobbering takes place.

When set to `false`, the classes will be flattened into a single directory. This essentially leaves the code use as it was before, although the generated sub-classes will now be put in their own file so they can be found, rather than added to the same file as the parent class. 

I also added RSpec snapshot testing to make this test easier to deal with. Also updated fig_tree to have a `with_config` method to make testing more succinct.

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## How Has This Been Tested?

- Unit tests

## Checklist:

cc: @iMacTia @DeeChau


DeeChau (approved):
I like the changes to use snapshots for rspecs, definitely cleaner than what we had implemented at first

Re: Snapshots + changes
the `### Secondary Schema Classes ###` are not aligned with the nested classes

e.g.

```rb
module Schemas
  ### Primary Schema Class ###
  # Autogenerated Schema for Record at com.my-namespace.MyNestedSchema
  class MyNestedSchema < Deimos::SchemaClass::Record

  ### Secondary Schema Classes ###
    # Autogenerated Schema for Record at com.my-namespace.MyNestedRecord
    class MyNestedRecord < Deimos::SchemaClass::Record
```

Otherwise the code changes + rspec updoots to switch between nested and non-nested generation looks great!

iMacTia (dismissed):
Fantastic work @dorner, thank you for taking the time to introduce this config and solve the autoloading issues 🙏 
I also welcome the introduction of `rspec-snapshot` which made it really easy to see what the end result would be after these changes.

Considering the breaking change, will this be released as v2.0?

If that's the case, can we use this chance to bump the minimum required Ruby version as well, or does your really old project also need these changes?
I'm asking because `zeitwerk` is a relatively new thing and I don't think it was around or is compatible with Ruby 2.3/2.4, hence this might be a really good change to pin your old project to version `< 2`, and release v2.0 with a more up-to-date minimum ruby version (ideally 2.7, but 2.6 might work as well 😄).

Also, before v2.0 is released, it would be great to double-check if there are any other breaking changes that might be worth adding while we're at it?

DeeChau (approved):
LGTM!!! 🚀

PR #163, opened 2022-09-08
Add `tombstone` method to schema classes

dorner:
This introduces the `tombstone` method to schema classes so that calling code doesn't have to know about the special "payload_key" value.


lib/generators/deimos/schema_class/templates/schema_record.rb.tt
```diff
@@ -62,7 +62,15 @@
     def namespace
       '<%= @current_schema.namespace %>'
     end
-
+<%- if @tombstone_assignment %>
+    def self.tombstone(key)
+      record = self.new
+<%- if @tombstone_assignment.present? -%>
+      <%= @tombstone_assignment %>
+<%- end -%>
```
angad-singh, line 70: Is this `if @tombstone_assignment` checking `false`/`true` and the next one checking for `nil`?

regenerate_test_schema_classes.rb
```diff
@@ -0,0 +1,68 @@
+#! /usr/bin/env ruby
+require 'action_controller/railtie'
```
angad-singh, line 2: Is this file just for testing?

lib/generators/deimos/schema_class/templates/schema_record.rb.tt
dorner: The first check is for `nil` and the second is for blank string.

angad-singh (approved):

PR #165, opened 2022-09-28
Add retries to DB poller and bypass "bad batches"

dorner:


lib/deimos/utils/db_poller.rb
```diff
@@ -143,15 +150,50 @@ def fetch_results(time_from, time_to)
...
+          Deimos.config.logger.error("Error publishing through DB Poller: #{e.message}")
+          sleep(0.5)
+          retry
+        rescue StandardError => e
+          if retries < @config.retries
+            retries += 1
+            sleep(0.5)
+            retry
+          else
+            Deimos.config.tracer&.set_error(span, e)
```
harsha-flipp, line 172: I was able to see the retry and tracer working with local setup. 
> Mock span 'deimos-db-poller' set an error: at .item_id expected type string, got null

Would you find it useful to log the error here in case they don't use datadog tracer?

harsha-flipp (dismissed):

PR #166, opened 2022-10-03
Tracer should be Tracing

rwilsonncsa:
# Pull Request Template

## Description

In Basic Configuration, "Tracer" is used where "Tracing" is meant.

## Type of change
Documentation typo.

## How Has This Been Tested?
Copied config entry as it was in the documentation got error, changed documentation to match what works.

## Checklist:
I just deleted the checklist because none seem to apply to doc change only.


dorner (approved):
Thanks so much for the help!

PR #167, opened 2022-10-25
Add the state_based mode for DB pollers.

dorner:
See README changes. This replaces the existing `DbPoller` class with a `DbPoller::Base`, `DbPoller::TimeBased` and `DbPoller::StateBased` classes.


dkharytonenko-blackbird (changes requested):
Hey, Daniel.
I got an error when starting db_poller
`undefined method `start!' for Deimos::Utils::DbPoller:Module`
I seems we need to call method `Deimos::Utils::DbPoller::Base.start!` in tasks/deimos.rake:33

dkharytonenko-blackbird (approved):

PR #170, opened 2022-11-15
GUILD-690: Initial draft of new Active Record Consumer Generator

jenchanflipp:
# Pull Request Template

## Description

Fixes # (issue)

## Type of change

## How Has This Been Tested?

- [ ] Test A
- [ ] Test B

## Checklist:


lib/generators/deimos/active_record_consumer/templates/migration.rb.tt
```diff
@@ -0,0 +1,30 @@
+class <%= migration_class_name %> < ActiveRecord::Migration<%= migration_version %>
```
dorner, line 1: Don't we already have this template file? We should be able to reuse it.

lib/generators/deimos/active_record_consumer_generator.rb
```diff
@@ -0,0 +1,151 @@
...
+        end
+
+        # Adds consumer config to config file.
+        # Defaults to deimos.rb if config_path arg is not specified.
+        def create_consumer_config
+          config_file = 'deimos.rb'
+          config_file_path = "#{initializer_path}/#{config_file}"
+          config_file_path = config_path if config_path.present?
+
+          if File.exist?(config_file_path)
```
dorner, line 52: We should validate that the config file exists if it's passed in. If not, we can create it.

lib/generators/deimos/active_record_consumer_generator.rb
```diff
@@ -0,0 +1,151 @@
...
+            config_template = File.expand_path(find_in_source_paths('config.rb'))
+            insert_into_file(config_file_path.to_s, CapturableERB.new(::File.binread(config_template)).result(binding))
```
dorner, line 54: Wouldn't this insert the whole file? We need a way to just insert the `consumer do` block.

lib/generators/deimos/active_record_consumer_generator.rb
```diff
@@ -0,0 +1,151 @@
...
+
+        # @return [String]
+        def model_name
+          table_name.underscore.singularize
+        end
+
+        # @return [String]
+        def consumer_name
+          "#{schema.classify}Consumer"
+        end
```
dorner, line 122: A lot of this is copied from the migration template. There should be a way to just call the migration generator from this generator rather than having the same code twice.

lib/generators/deimos/active_record_consumer_generator.rb
```diff
@@ -0,0 +1,151 @@
...
+            paths = Rails.application.config.paths['config/initializers']
+            paths.respond_to?(:to_ary) ? paths.to_ary.first : paths.to_a.first
+          else
+            'config/initializers'
+          end
+        end
+
+        # @return [String] Returns the name of the first field in the schema, as the key
+        def key_field
+          fields.first.name
```
dorner, line 136: This is almost definitely not going to be the right approach. It probably makes more sense to just have the key config specified as part of the input. We'll also need the key schema (if there is one) when generating the schema classes.

lib/generators/deimos/active_record_consumer_generator.rb
```diff
@@ -0,0 +1,151 @@
...
+            config_template = File.expand_path(find_in_source_paths('config.rb'))
+            insert_into_file(config_file_path.to_s, CapturableERB.new(::File.binread(config_template)).result(binding))
+          else
+            template('config.rb', config_file_path.to_s)
+          end
+        end
+
+        # Generates schema classes
+        def create_deimos_schema_class
+          Deimos::Generators::SchemaClassGenerator.start
```
dorner, line 62: There should be a way to generate just the one schema class we care about.

dorner (changes requested):

dorner:
This is a great first pass! I do think that having an interactive process will make more sense, especially because right now the templates are making a bunch of assumptions as to what needs to be filled in. Allowing the user to specify that will help a lot so they don't have to go into each of these files and edit them.

lib/generators/deimos/active_record_consumer/templates/migration.rb.tt
jenchanflipp: Yes - better idea to reuse the existing code - will fix this.

lib/generators/deimos/active_record_consumer_generator.rb
jenchanflipp: When I tested it by passing in a config file path that did not already exist, it did go ahead and create that config file.

dorner: Sorry, that was unclear. We should only create it if no path was specified. If a path was specified and doesn't exist, we should assume that means that they made a mistake and raise an error.

Actually, I'm not sure we need to create it at all. If we want to auto-create a config file, it probably would be in an installer or something, but I don't think we have a use case for that. So maybe I'd just say don't create anything, and throw an error if the file doesn't exist.

lib/generators/deimos/active_record_consumer_generator.rb
jenchanflipp: I considered extending from the migration template but decided it wasn't the right approach. Will find a better way!

lib/generators/deimos/active_record_consumer_generator.rb
jenchanflipp: Ok, yes makes sense to include key config as an input.

In terms of the key schema, how does that work? I noticed that the schema glasses generator doesn't take in any arguments. So, if the user has a key schema for the new consumer, do they just have to place it in the SCHEMA_ROOT prior to running the ActiveRecordConsumer generator? And then what will be the difference in output? Will the generated schema classes be different?

dorner: The generator reads the configuration which already knows what the key schema is.

The generator has a method `generate_classes(schema_name, namespace, key_config)` which you should probably be using.

lib/generators/deimos/active_record_consumer_generator.rb
dorner: The one that was passed in as an input.

dorner:
I'd look through the template and generator files to see what assumptions are being made in the output, and try to add those as inputs.

PR #171, opened 2022-12-16
CCOL-1590: Bulk Import handles associations

harsha-flipp:
## Description

ActiveRecord Consumer has capability to do bulk imports with MySQL. This change adds to that capability by handling associations for the activerecord and bulk save them.
- The application code has to set two parameters --
1. validation context(optional) - defaults to nil. Helps apply filtered validations otherwise uses all validations in the default scope
2. assocation_list - Helps identify which associations deimos should handle. There could be multiple associations to an ActiveRecord model.
Books - Say, books have id and language_id which is association to another table.
Books could have association on authors which use `id` and langugages which use `language_id`. Using reflections I cannot filter which associations are relevant for filling ids and save associations.

Fixes # (issue)
No issue filed

## Type of change

- [ X ] New feature (non-breaking change which adds functionality)
- [ X ] This change requires a documentation update

## How Has This Been Tested?

- [ ] Test A
- [ ] Test B

## Checklist: TODO


lib/deimos/active_record_consume/batch_consumption.rb
```diff
@@ -88,36 +88,73 @@ def update_database(messages)
...
+
+        save_records_to_database(@klass, key_cols, valid_upserts)
+
+        import_associations(valid_upserts, key_cols) unless @association_list.empty?
+      end
+
+      def save_records_to_database(record_class, key_cols, records)
+        columns = :all
+        if record_class.respond_to?(:bulk_import_columns)
+          columns = record_class.bulk_import_columns
```
dorner, line 106: I'd want this to be a configuration for the consumer, not the record class. Let's try to keep the model class itself clean. Probably the easiest way is just to have a method on the consumer that takes a class and returns the columns.

lib/deimos/active_record_consume/batch_consumption.rb
```diff
@@ -163,6 +200,41 @@ def compact_messages(batch)
...
+
+      # Filters list of Active Records by applying active record validations. Optionally, validation_context can
+      # be set in ActiveRecordConsumer to override the default `nil` context.
+      # Tip: Add validates_associated in ActiveRecord model to validate associated_models
+      # Optionally inherit this method and apply more filters in the application code
+      # @param records Array<ActiveRecord> - List of active records which will be subjected to model validations
+      # @return valid Array<ActiveRecord> - Subset of records that passed the model validations
+      def filter_models(records)
+        valid, invalid = records.partition do |p|
+          p.valid?(@validation_context)
```
dorner, line 229: I feel like adding a context here is actually not a common thing to do - usually things are either valid or they're not. I'd leave out the context here and override this in item-platform to add the context. It's a small enough method to be safe.

lib/deimos/active_record_consume/batch_consumption.rb
```diff
@@ -88,36 +88,73 @@ def update_database(messages)
...
+      # config for associations
+      # bulk_import_id is required only if handle_associations is set to true
+      # alternatively set a config to add which associations need to be considered
+      # There could be many associations not relevant to master table. For example, products and merchant association
+      def import_associations(entities)
+        associations = @klass.reflect_on_all_associations.select { |assoc| @association_list.include?(assoc.name) }
+
+        # create DB migration for bulk insert id. I did it manually in item-platform
+        table_by_bulk_import_id =
+          @klass.where(bulk_import_id: entities.map(&:bulk_import_id)).
```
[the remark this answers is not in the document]
harsha-flipp: `bulk_import_id` is not required if they don't want to handle associations. it's something user has to define as a unique ID during build_models phase.
I am trying to understand how to prompt user to run a migration to create a column and index on this?

I will take a first pass on what I would do. 
Should I do something like `self.klass.columns.include?(@config.bulk_import_id_column)` in this method? 
If the column does not exist, raise an exception with information, say, "Create DB Migration on primary table and set bulk_import_id in build_models. Overriding build_models is mandatory if you want to handle associations"

Some message like this. 🤔

lib/deimos/active_record_consume/batch_consumption.rb
dorner: the nice thing about the method is that it can be whatever you want it to be as long as it returns an array of column names :)

lib/deimos/active_record_consume/batch_consumption.rb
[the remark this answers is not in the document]
dorner: Yeah something like that makes sense. I think we can treat `bulk_import_id` as - if you have it in your model, then we fully control it. So we can set it and read it back out. If you don't have it, but you indicated you want us to save your associations, then we can raise an error and say "please add this column by running `deimos install:bulk_import` or something similar.

harsha-flipp:
Existing tests pass. I need to add unit tests and address a couple of review comments.
 
 I will open it up for review after I address these.

lib/deimos/active_record_consume/batch_consumption.rb
```diff
@@ -163,6 +200,41 @@ def compact_messages(batch)
 
         batch.reverse.uniq(&:key).reverse!
       end
+
+      # Turns Kafka payload into ActiveRecord Objects by mapping relevant fields
+      # Override this method to build object and associations with message payload
+      # @param messages [Array<Deimos::Message>] the array of deimos messages in batch mode
+      # @return [Array<ActiveRecord>] Array of ActiveRecord objects
+      def build_models(messages)
```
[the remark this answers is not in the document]
harsha-flipp: Renamed.

lib/deimos/active_record_consume/batch_consumption.rb
harsha-flipp: Done. I tested this in RSpecs as well

lib/deimos/active_record_consume/batch_consumption.rb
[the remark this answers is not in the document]
harsha-flipp: Done. This was tricky one. I ended up adding this to ActiveRecord class properties
e.g.,
```
class WidgetConsumer <  Deimos::ActiveRecordConsumer
          record_class Widget
          association_list :detail
          bulk_import_id_column :custom_id
end
```

I think this would not live in flipp_ruby_kafka kind of configuration level similar to record_class and association_list.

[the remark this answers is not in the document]
harsha-flipp: I called the generator as `rails g deimos:bulk_import_id:setup`. Let me know if this is not a great choice for the generator. Will change the name.

lib/deimos/active_record_consume/batch_consumption.rb
harsha-flipp: Removed.

lib/deimos/active_record_consume/batch_consumption.rb
```diff
@@ -88,36 +88,73 @@ def update_database(messages)
...
+            select(:id, :bulk_import_id).
+            index_by(&:bulk_import_id)
+        # update IDs in master table
+        entities.each { |entity| entity.id = table_by_bulk_import_id[entity.bulk_import_id].id }
+        associations.each do |assoc|
+          details = entities.map { |master|
+            details = master.send(assoc.name)
+            unless details.is_a?(Enumerable)
+              details = [details]
+            end
```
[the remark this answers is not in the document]
harsha-flipp: Done

lib/deimos/active_record_consumer.rb
```diff
@@ -30,6 +30,17 @@ def record_class(klass)
...
+      def validation_context(context)
+        config[:validation_context] = context
+      end
+
+      def association_list(associations)
+        if associations.is_a?(Array)
+          config[:association_list] = associations
+        else
+          config[:association_list] = [associations]
+        end
```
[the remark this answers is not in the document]
harsha-flipp: Done

spec/active_record_batch_consumer_mysql_spec.rb
```diff
@@ -0,0 +1,219 @@
+# frozen_string_literal: true
+
```
harsha-flipp, line 2: The existing test runs with SQLite with `bundle exec rspec`.

The feature of association_list is with MySQL so I had to write this `:integration` test for testing mysql changes.
Will write up a ticket to setup and run integration tests on deimos. 
For Non-MySQL DBs, we raise NotImplementedError which has been tested in the other RSpec.

lib/deimos/active_record_consume/batch_consumption.rb
```diff
@@ -88,36 +88,85 @@ def update_database(messages)
...
+          select(:id, @bulk_import_id_column).
+          index_by { |e| e[@bulk_import_id_column] }
+        # update IDs in master table
+        entities.each { |entity| entity.id = table_by_bulk_import_id[entity[@bulk_import_id_column]].id }
+        associations.each do |assoc|
+          details = entities.map { |master|
+            details = Array(master.send(assoc.name))
+            details.each { |d| d.send("#{assoc.send(:foreign_key)}=", master.id) }
+            details
+          }.flatten
```
dorner, line 159: I think I need some more comments here... I _think_ I know what this is doing but I'm having a hard time following with all the configurable values.

lib/deimos/active_record_consume/batch_consumption.rb
```diff
@@ -88,36 +88,85 @@ def update_database(messages)
...
+
+          # is there a way to replace key_cols to _ to mention that we dont care
```
dorner, line 161: Only when defining a function, not calling it :)

lib/deimos/active_record_consume/batch_consumption.rb
```diff
@@ -163,6 +212,40 @@ def compact_messages(batch)
...
         batch.reverse.uniq(&:key).reverse!
       end
+
+      # Turns Kafka payload into ActiveRecord Objects by mapping relevant fields
+      # Override this method to build object and associations with message payload
+      # @param messages [Array<Deimos::Message>] the array of deimos messages in batch mode
+      # @return [Array<ActiveRecord>] Array of ActiveRecord objects
+      def build_records(messages)
+        # This method must be overriden when association_list is configured
+        raise NotImplementedError unless @association_list.blank?
```
dorner, line 222: Probably should have a better error message here. Also, I learned that `NotImplementedError` is actually the wrong thing to use here. In other repos I had a `MissingImplementationError` - if we don't have one, we probably should create one and move any other `NotImplementedError` cases to use it.

lib/deimos/active_record_consume/batch_consumption.rb
```diff
@@ -163,6 +212,40 @@ def compact_messages(batch)
...
+      # Tip: Add validates_associated in ActiveRecord model to validate associated models
+      # Optionally inherit this method and apply more filters in the application code
+      # The default implementation throws ActiveRecord::RecordInvalid by default
+      # @param records Array<ActiveRecord> - List of active records which will be subjected to model validations
+      # @return valid Array<ActiveRecord> - Subset of records that passed the model validations
+      def filter_records(records)
+        records.each(&:validate!)
+      end
+
+      def my_sql_adapter?
```
dorner, line 246: Should be `mysql_adapter?`

README.md
```diff
@@ -353,6 +353,45 @@ class MyBatchConsumer < Deimos::Consumer
...
+    end
+  end
+  
+  def bulk_import_columns(klass)
+    case klass
+    when User
+      key_cols, cols = [["first_name"], all_cols - timestamps]
+    when Image
+      key_cols, cols = [["image_url", "image_name"], all_cols - timestamps - id]
+    end
```
dorner, line 391: 1.  Can we have a sane default which we override in the consumer instead of checking to see if the method is defined?
2.  Can we split this out into 2 methods, one which calculates key columns and one import columns?
3. How does this interact with the existing `record_key` method? Is there a way to reuse that?

dorner (changes requested):
@harsha-flipp I'm finding it a bit hard to follow the code... might need to be better organized or commented. But I added some comments based on what I think it's doing.

The important thing IMO is to try (as much as possible) to unify the flow between non-association and association. I know we have a number of dimensions now (are we saving associations? Is it a bulk or per-message consumption?) which is why trying to reduce flow differences is important. It might not be all that possible, but I think we can do a bit better.

lib/deimos/active_record_consume/batch_consumption.rb
harsha-flipp: I refactored this method a bit and moved that dense block in `import_assocations` method based on functionality. Please take a look when you get a chance.

lib/deimos/active_record_consume/batch_consumption.rb
harsha-flipp: Thank you. I wrote it as a reminder and removed that line.

lib/deimos/active_record_consume/batch_consumption.rb
harsha-flipp: I made MissingImplementationError in deimos/lib/exceptions.rb. 
There are very many NotImplementedError instances in Deimos. To avoid too many files changes, after this review is complete, I can add it as a separate commit in this PR.

lib/deimos/active_record_consume/batch_consumption.rb
harsha-flipp: Done.

README.md
harsha-flipp: Great questions.
1 and 2 helped me refactor the flow between associations and entities. `save_records_to_database` is used by both and if-else statements are removed to unify the flow. 

3. I had a chat about this on slack and we decided to change method signature a bit.
`def key_columns(messages, klass)`
Initially I thought of raising errors when using key_columns with `association_list` like
```
def key_columns...
   raise MissingImplementationError unless @association_list.empty?
end
```
One downside is that the application cannot call a `super` if I enforce that. I decided to not raise errors.
In the documentation, we can add that `key_columns(messages, record_class)` and `columns(record_class)` should be overriden because we don't know anything about the associated table.

lib/deimos/active_record_consume/batch_consumption.rb
```diff
@@ -86,38 +87,63 @@ def update_database(messages)
...
+              # Get associated `has_one` or `has_many` records for each entity
+              details = Array(entity.send(assoc.name))
+              # Set IDS from master to each of the records in `has_one` or `has_many` relation
+              details.each { |d| d.send("#{assoc.send(:foreign_key)}=", entity.id) }
+              details
+            }.flatten
 
-        @klass.import!(upserts, options)
+            columns = key_columns(nil, assoc.klass)
+            save_records_to_database(assoc.klass, columns, details) unless details.empty?
```
dorner, line 145: Prefer positive conditions - `if details.any?`

lib/deimos/active_record_consume/batch_consumption.rb
```diff
@@ -145,15 +171,27 @@ def deleted_query(records)
       # Get the set of attribute names that uniquely identify messages in the
       # batch. Requires at least one record.
       # @param records [Array<Message>] Non-empty list of messages.
+      # @param _klass [ActiveRecord::Class] Class Name can be used to fetch columns
       # @return [Array<String>] List of attribute names.
       # @raise If records is empty.
-      def key_columns(records)
+      def key_columns(records, _klass)
```
dorner, line 177: I think we can change this method so that it works in both cases and still allows `super`:

```ruby
def key_columns(records, klass)
  if klass.nil? # check records
    raise 'Cannot determine key from empty batch' if records.empty?

        first_key = records.first.key
        record_key(first_key).keys
  else
     raise 'Must implement key_columns method for associations!
  end
end
```

In this case:
* Calling it with no class passed in (default case) will do the current default behavior. You can call `super` here to get the current default behavior.
* Calling it with a subclass will have no default behavior and will raise an error. You can't call `super` here anyway because we don't have any defined behavior.

lib/deimos/active_record_consume/batch_consumption.rb
```diff
@@ -86,38 +87,63 @@ def update_database(messages)
...
+      # @association_list configured on the consumer helps identify the ones required to be saved.
+      def import_associations(entities)
+        _validate_associations(entities)
+        _fill_primary_key_on_entities(entities)
+
+        # Select associations from config parameter association_list and
+        # fill id to associated_objects foreign_key column
+        @klass.reflect_on_all_associations.select { |assoc| @association_list.include?(assoc.name) }.
+          each do |assoc|
+            details = entities.map { |entity|
```
dorner, line 136: Can we call this something like `sub_records` rather than `details`?

dorner (changes requested):
Much easier to follow now! A couple of remaining comments.

lib/deimos/active_record_consume/batch_consumption.rb
harsha-flipp: Done.

lib/deimos/active_record_consume/batch_consumption.rb
harsha-flipp: Done.

dorner (approved):

dorner:
Looks good! Have you tested this with item-platform @harsha-flipp ?

dorner (approved):

PR #179, opened 2023-03-27
PDA-112: Removing associations on update

randommancer:
# Pull Request Template

## Description

Removing associations when updating.

Fixes # (issue)

Bug / Feature

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## How Has This Been Tested?

- [x] Executed and add a new test on `active_record_batch_consumer_mysql_spec.rb`

## Checklist:

### DIFF
--- a/lib/deimos/active_record_consume/batch_consumption.rb
+++ b/lib/deimos/active_record_consume/batch_consumption.rb
@@ -89,6 +89,8 @@ module Deimos
       def upsert_records(messages)
         key_cols = key_columns(messages, @klass)
 
+        # Remove associations before update
+        remove_associations(messages)
         # Create ActiveRecord Models with payload + key attributes
         upserts = build_records(messages)
         # If overridden record_attributes indicated no record, skip
@@ -100,6 +102,19 @@ module Deimos
 
         save_records_to_database(@klass, key_cols, valid_upserts)
         import_associations(valid_upserts) unless @association_list.blank?
+      end
+
+      def remove_associations(messages)
+        keys = messages.map(&:key)
+        query = @klass.unscoped.where(:id => keys)
+        if query.exists?
+          entities = query.take(messages.length)
+          @klass.reflect_on_all_associations.select { |assoc| @association_list.include?(assoc.name) }.
+            each do |assoc|
+            assoc_values = entities.map { |entity| entity.send(assoc.active_record_primary_key) }
+            assoc.klass.unscoped.where(assoc.foreign_key => assoc_values).delete_all
+          end
+        end
       end
 
       def save_records_to_database(record_class, key_cols, records)
--- a/lib/deimos/active_record_consume/batch_consumption.rb
+++ b/lib/deimos/active_record_consume/batch_consumption.rb
@@ -117,14 +117,16 @@ module Deimos
       end
 
       def remove_associations(assoc, keys)
-        foreign_keys = keys.map { |keys| keys[assoc.foreign_key] }.compact
-        associated_keys = keys.map { |keys| keys[associated_column] }.compact
-        if foreign_keys.length > 0 && associated_keys.length > 0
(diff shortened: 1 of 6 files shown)


lib/deimos/active_record_consume/batch_consumption.rb
```diff
@@ -102,6 +104,14 @@ def upsert_records(messages)
         import_associations(valid_upserts) unless @association_list.blank?
       end
 
+      def remove_associations(messages)
```
dorner, line 107: So thinking about this a bit more... this is probably unnecessary in every case because this is really just to cover the case where a previously existing association is no longer in the message. It probably makes sense to move this functionality to `import_associations`with the following basic structure:

*   Get all association keys from the payload
*   Do the import
*   Delete from the association where the key is NOT in the list of keys we got

Otherwise we'll be deleting and re-creating data every time.

dorner (changes requested):

lib/deimos/active_record_consume/batch_consumption.rb
```diff
@@ -116,6 +116,23 @@ def upsert_records(messages)
         import_associations(valid_upserts) unless @association_list.blank?
       end
 
+      def remove_associations(assoc, keys)
+        foreign_keys = keys.map { |key| key[assoc.foreign_key] }.compact
+        associated_keys = keys.map { |key| key[associated_column] }.compact
+        if foreign_keys.length.zero? && associated_keys.length.zero?
```
dorner, line 122: More Rubyish:

`if foreign_keys.empty? && associated_keys.empty?`

lib/deimos/active_record_consume/batch_consumption.rb
```diff
@@ -116,6 +116,23 @@ def upsert_records(messages)
...
+          return
+        end
+
+        assoc.klass.unscoped.where.not(
+          assoc.foreign_key => foreign_keys,
+          associated_column => associated_keys
+        ).delete_all
```
dorner, line 129: Oof this isn't going to work. Imagine if we have the following:

FK1->AC1  
FK2->AC1  
FK3->AC3  
FK4->AC4

and we have the following messages coming in:  
  
FK1->AC1

What this query is saying is:

"if a record doesn't have any of the given foreign keys (FK1) and doesn't have any of the given association keys (AC1) delete it"

This would delete every row in the database besides the first one! :grin: 

The only way we could do this would be to do a query one FK (i.e. one offer\_id) at a time, or to join it together in a giant WHERE clause... neither of which are great :thinking: 

I apologize for the fuzzy thinking. We'll need to think this through a bit more carefully.

lib/deimos/active_record_consume/batch_consumption.rb
```diff
@@ -142,24 +159,74 @@ def save_records_to_database(record_class, key_cols, records)
...
           each do |assoc|
+
+            records_to_remove = []
             sub_records = entities.map { |entity|
               # Get associated `has_one` or `has_many` records for each entity
               sub_records = Array(entity.send(assoc.name))
               # Set IDS from master to each of the records in `has_one` or `has_many` relation
+
+              new_records = sub_records.map { |record| record.attributes['id'] == nil ? record : nil }.compact
+              old_records = sub_records.map { |record| record.attributes['id'] != nil ? record : nil }.compact
```
dorner, line 176: This assumes that the records coming in would have an ID. In the offer/product use case, there will **never** be an ID because the downstream system doesn't "own" the data, it's only requesting you to update the data that you own. I don't think this will work either. :disappointed: I think there is a solution we can use - will chat about it once PDA-114 is done.

randommancer: Sounds great, we can chat about that after PDA-114

randommancer:
Reworked on another PR

PR #181, opened 2023-04-04
[CCOL-1874]: Add functionality to allow DB poller to be able to publish a single poll query to multiple topics.

MichaelCondo:
# Pull Request Template

## Description

This PR is to add functionality to allow state based db pollers to not be tied to a single producer and allow the publishing to multiple kafka topics for a single poll query. 

Fixes # (issue)

## Type of change

- New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

- [ ] TODO
- [ ] TODO

## Checklist:


lib/deimos/config/configuration.rb
```diff
@@ -481,6 +481,10 @@ def self.configure_producer_or_consumer(kafka_config)
       setting :published_state
       # Value to set the state_column to if publishing fails - state-based only.
       setting :failed_state
+
+      # Inherited state based poller class name to use for publishing
```
dorner, line 485: Actually it shouldn't need to be state based - any subclass should work.

lib/deimos/utils/db_poller/base.rb
```diff
@@ -21,17 +21,39 @@ class Base
         # @return [Hash]
         attr_reader :config
 
+        # Method to define producers if a single poller needs to publish to multiple topics.
+        # Producer classes should be constantized
+        # @return [Array<Producer>]
+        def self.multi_producers
```
dorner, line 27: This can just be `self.producers`.

lib/deimos/utils/db_poller/base.rb
```diff
@@ -21,17 +21,39 @@ class Base
...
-          end
-          unless @producer < Deimos::ActiveRecordProducer
-            raise "Class #{@producer.class.name} is not an ActiveRecordProducer!"
+            @poller_class = @config.poller_class.present? ? @config.poller_class.constantize : nil
+            if self.class.multi_producers.empty?
+              if @config.producer_class.nil?
+                raise 'No producers have been set for this DB poller!'
+              end
+
+              @producer = @config.producer_class.constantize
```
dorner, line 43: There's no need for this instance variable any more. You're going to want to validate all the producers that have been configured.

lib/deimos/utils/db_poller/base.rb
```diff
@@ -21,17 +21,39 @@ class Base
...
+              @producers << @producer
+              unless @producer < Deimos::ActiveRecordProducer
+                raise "Class #{@producer.class.name} is not an ActiveRecordProducer!"
+              end
+            else
+              @producers.append(self.class.multi_producers).flatten!.each do |producer|
```
dorner, line 49: Is there a reason we need to flatten it? And why would we need to append the producer list again when we already have it?

lib/deimos/utils/db_poller/base.rb
```diff
@@ -98,13 +120,16 @@ def process_updates
...
-            )
-            process_batch(batch)
-            Deimos.config.tracer&.finish(span)
+            @producers.each do |producer|
+              span = Deimos.config.tracer&.start(
+                'deimos-db-poller',
+                resource: producer.class.name.gsub('::', '-')
+              )
+              process_batch(producer, batch)
+              Deimos.config.tracer&.finish(span)
```
dorner, line 131: I think you should have just one span - it should be set to the poller class rather than the producer class if the poller class is a subclass. Something like 

```ruby
resource_class = self.producers.any? ? self : @producer 
```

Probably best to set this resource_class during initialization since you're in the middle of that logic then anyway.

lib/deimos/utils/db_poller/state_based.rb
```diff
@@ -26,12 +26,16 @@ def process_updates
             success = process_batch_with_span(batch, status)
             finalize_batch(batch, success)
           end
-          Deimos.config.logger.info("Poll #{@producer.topic} complete (#{status.report}")
+          Deimos.config.logger.info("Poll #{@producers.map(&:topic)} complete (#{status.report}")
```
dorner, line 29: This probably should be a method on the base poller class e.g.

```
def log_identifier
  "#{self.class.name}: #{@producers.map(&:topic)}"
end  
```

Whatever we end up doing with the aforementioned resource_class can help this.

lib/deimos/utils/db_poller/state_based.rb
```diff
@@ -26,12 +26,16 @@ def process_updates
...
         end
 
         # @return [ActiveRecord::Relation]
         def fetch_results
-          @producer.poll_query.limit(BATCH_SIZE).order(@config.timestamp_column)
+          if @poller_class.nil?
+            @producer.poll_query.limit(BATCH_SIZE).order(@config.timestamp_column)
+          else
+            @poller_class.poll_query.limit(BATCH_SIZE).order(@config.timestamp_column)
```
dorner, line 37: Using `resource_class` can help here too.

dorner (changes requested):

lib/deimos/config/configuration.rb
MichaelCondo: @dorner Sounds good. I guess in terms of the time_based poller class, I would also need to make the same changes I did to the state_based poller class (i.e. updated the fetch_results method)?

lib/deimos/utils/db_poller/base.rb
MichaelCondo: @dorner The reason why I appended the producer list to the @producers instance variable is so that I could use the @producers instance variable throughout the base file (as I set it in both the multi producer and single producer configurations), especially when doing the Deimos config resource tracer spans.

lib/deimos/utils/db_poller/base.rb
MichaelCondo: @dorner I noticed you asked me to remove the @producer instance variable above, but I see reference to it on this comment. Should we still keep that variable?

dorner: No, you don't need it - I mean that when you're looking at setting up the producers, you can also set the resource\_class there as well.

lib/deimos/utils/db_poller/base.rb
dorner: But the point is that either `self.producers` returns a value, or `@producer` is given. There's no reason to combine the two since only one of them will have a value.

lib/deimos/utils/db_poller/base.rb
MichaelCondo: Ah okay I see. So instead of storing the @producer variable, I can just look at the config.producer_class directly 👍

lib/deimos/utils/db_poller/base.rb
```diff
@@ -65,12 +85,12 @@ def start
...
         # @return [void]
         def retrieve_poll_info
-          @info = Deimos::PollInfo.find_by_producer(@config.producer_class) || create_poll_info
+          @info = Deimos::PollInfo.find_by_producer(@resource_class.to_s) || create_poll_info
         end
 
         # @return [Deimos::PollInfo]
         def create_poll_info
-          Deimos::PollInfo.create!(producer: @config.producer_class, last_sent: Time.new(0))
+          Deimos::PollInfo.create!(producer: @resource_class.to_s, last_sent: Time.new(0))
```
MichaelCondo, line 93: @dorner I put this in just for now so I could test it with item-platform locally to verify it works. I wanted to get your thoughts though on if maybe we should rename the `producer` attribute to be something else that makes more sense considering it can be either a single producer class name, or the poller class name.

lib/deimos/utils/db_poller/base.rb
```diff
@@ -21,17 +21,37 @@ class Base
...
+
+            if @config.poller_class.nil?
+              unless @resource_class < Deimos::ActiveRecordProducer
+                raise "Class #{@resource_class.class.name} is not an ActiveRecordProducer!"
+              end
+            else
+              @resource_class.producers.each do |producer|
+                unless producer < Deimos::ActiveRecordProducer
+                  raise "Class #{producer.class.name} is not an ActiveRecordProducer!"
+                end
```
dorner, line 50: This is some copied code - can we extract this to a method? E.g. `validate_producer_class`?

lib/deimos/utils/db_poller/base.rb
```diff
@@ -21,17 +21,37 @@ class Base
...
+              end
+            end
+          rescue NameError => e
+            raise e.message.to_s
```
dorner, line 54: Is there a reason we're just re-raising the error here?

lib/deimos/utils/db_poller/base.rb
dorner: Yes, I think that makes a lot of sense.

lib/deimos/utils/db_poller/base.rb
```diff
@@ -128,10 +148,28 @@ def process_batch_with_span(batch, status)
...
+        end
+
+        # Configure log identifier and messages to be used in subclasses
+        # # @return [String]
+        def log_identifier
+          if @config.producer_class.present?
+            "#{self.class.name}: #{@resource_class.topic}"
+          else
+            "#{self.class.name}: #{@resource_class.producers.map(&:topic)}"
+          end
```
dorner, line 172: I think there's something fundamental weird here... `@resource_class` should be `self` if we're using `poller_class` instead of `producer_class`. This is the base class, so we should just be able to reference `self.producers`.

Maybe we need an intermediate method, e.g.

```ruby
def producer_classes
  return self.producers self.producers.any?
  [@producer_class]
end
```

and then one more:
```ruby
def self.poller_class
  @producer_class || self
end
```

There should be very few `if` statements when we have those.

dorner (changes requested):

lib/deimos/utils/db_poller/base.rb
MichaelCondo: @dorner Yes so the reason why I changed the code to use these if statements: `if @config.producer_class.present?` in order to determine if it should choose the producer or poller path, is because when I was testing using item-platform, I noticed that when I had `self.producers`, the self was the StateBased poller class, which would return an empty array since the producers method is actually setup on the inherited poller class. 
Not too sure if I am making sense or not.

dorner: If that's the case, then we're missing a piece - [https://github.com/flipp-oss/deimos/blob/master/lib/deimos/utils/db_poller.rb#L15](https://github.com/flipp-oss/deimos/blob/master/lib/deimos/utils/db_poller.rb#L15) - we need to instantiate the correct class here rather than always use the parent class.

lib/deimos/utils/db_poller/base.rb
MichaelCondo: I guess the reason I changed it was because before it was `raise "Class #{@config.producer_class} not found!"`, but now it could either be `@config.producer_class` or `@config.poller_class`, and if it fails when one of them get `constantize` called on it, I guess we would need to check which one is populated since at this stage if it fails, `@resource_class` wouldn't have even got populated.

dorner: If we fix having to deal with the `if` statements, this becomes a lot easier :) You can pass whatever class to the validation method and have it handle this.

lib/deimos/utils/db_poller/base.rb
MichaelCondo: @dorner In order to do this, would I just need to update the template files in the generator directory here:
`lib/generators/deimos/db_poller/templates/migration`
`lib/generators/deimos/db_poller/templates/rails3_migration`
And rename the producer attribute in the db migration? Or would I need make these alterations somewhere else?

lib/deimos/utils/db_poller/base.rb
MichaelCondo: Sounds good 👍

lib/deimos/utils/db_poller/base.rb
dorner: I forgot this was a DB column. Hmm... I'm not sure it's worth having to deal with changing existing applications in order to do this. Let's leave it.

MichaelCondo: Okay sounds good.

lib/deimos/utils/db_poller/base.rb
MichaelCondo: @dorner This makes sense. I guess I can adjust the class_for_config method here: https://github.com/flipp-oss/deimos/blob/master/lib/deimos/utils/db_poller.rb#L26
and first check if poller_class is in the config and instantiate that class, and if not, then continue with the code already in that method.

lib/deimos/utils/db_poller/base.rb
```diff
@@ -101,7 +113,7 @@ def process_batch_with_span(batch, status)
           begin
             span = Deimos.config.tracer&.start(
               'deimos-db-poller',
-              resource: @producer.class.name.gsub('::', '-')
+              resource: @resource_class.class.name.gsub('::', '-')
```
dorner, line 116: `@resource_class` is already a class, so wouldn't this always just show `Class`?

lib/deimos/utils/db_poller/base.rb
```diff
@@ -128,10 +140,35 @@ def process_batch_with_span(batch, status)
...
-          @producer.send_events(batch)
+          producer_classes.each do |producer|
+            producer.send_events(batch)
+          end
+        end
+
+        # Configure log identifier and messages to be used in subclasses
+        # @return [String]
+        def log_identifier
+          "#{self.class.name}: #{producer_classes.map(&:topic)}"
```
dorner, line 155: Shouldn't this be `@resource_class.name`?

lib/deimos/utils/db_poller/time_based.rb
```diff
@@ -44,20 +44,24 @@ def process_updates
...
         # @param time_from [ActiveSupport::TimeWithZone]
         # @param time_to [ActiveSupport::TimeWithZone]
         # @return [ActiveRecord::Relation]
         def fetch_results(time_from, time_to)
-          id = @producer.config[:record_class].primary_key
+          id = if @config.producer_class.present?
+                 @resource_class.config[:record_class].primary_key
+               else
+                 @resource_class.producers.first.config[:record_class].primary_key
+               end
```
dorner, line 58: You should be able to replace this with just

`@resource_class.producer_classes.first.config[:record_class].primary_key`

This should always return the right field.

dorner (changes requested):
Getting a lot closer @MichaelCondo !

lib/deimos/utils/db_poller/time_based.rb
MichaelCondo: @dorner Would this work considering that the `@resource_class` variable can be set to the constantized producer class as set in the initialize method:
`@resource_class = self.class.producers.any? ? self.class : @config.producer_class.constantize`
and the producer class doesn't have the `producer_classes` method since that method lives in the DbPoller class?

dorner: Sorry, my bad - should be `self.producer_classes`

MichaelCondo: No worries. Sounds good 👍

lib/deimos/utils/db_poller/base.rb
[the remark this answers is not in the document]
dorner: This still needs to be changed - `.class` needs to be removed

dorner (changes requested):
I think this is the last thing left!

lib/deimos/utils/db_poller/base.rb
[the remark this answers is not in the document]
MichaelCondo: @dorner I think this is from an old commit. The recent commit I just pushed should have that change:
https://github.com/flipp-oss/deimos/pull/181/files#diff-886ea5bd7b630e2f3469c2ac376642c8ddf31f8c4ca5ec73c25caef4e02d22e3R116

dorner (approved):

dorner:
Nice - thanks so much for the feature!

PR #200, opened 2023-08-24
Batch ulid

harsha-flipp:
# Pull Request Template

## Description

Fixes # (issue)

## Type of change

## How Has This Been Tested?

- [ ] Test A
- [ ] Test B

## Checklist:

### DIFF
--- a/lib/deimos/active_record_consume/batch_record.rb
+++ b/lib/deimos/active_record_consume/batch_record.rb
@@ -26,7 +26,7 @@ module Deimos
         @klass = klass
         if bulk_import_column
           self.bulk_import_column = bulk_import_column
-          self.bulk_import_id = SecureRandom.uuid
+          self.bulk_import_id = ULID.generate
           attributes[bulk_import_column] = bulk_import_id
         end
         attributes = attributes.with_indifferent_access
--- a/lib/deimos/active_record_consume/batch_record.rb
+++ b/lib/deimos/active_record_consume/batch_record.rb
@@ -26,7 +26,11 @@ module Deimos
         @klass = klass
         if bulk_import_column
           self.bulk_import_column = bulk_import_column
-          self.bulk_import_id = ULID.generate
+          self.bulk_import_id = if klass.method_defined?(:generate_unique_id)
+                                  klass.generate_unique_id
+                                else
+                                  SecureRandom.uuid
+                                end
           attributes[bulk_import_column] = bulk_import_id
         end
         attributes = attributes.with_indifferent_access
--- a/lib/deimos/active_record_consume/batch_consumption.rb
+++ b/lib/deimos/active_record_consume/batch_consumption.rb
@@ -22,7 +22,7 @@ module Deimos
       # they are split
       # @param payloads [Array<Hash,Deimos::SchemaClass::Record>] Decoded payloads
       # @param metadata [Hash] Information about batch, including keys.
-      # @return [void]
+      # @return [Array<ActiveRecord::Base>]
       def consume_batch(payloads, metadata)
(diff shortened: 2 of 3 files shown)


lib/deimos/active_record_consume/batch_record.rb
```diff
@@ -26,7 +26,11 @@ def initialize(klass:, attributes:, bulk_import_column: nil)
         @klass = klass
         if bulk_import_column
           self.bulk_import_column = bulk_import_column
-          self.bulk_import_id = SecureRandom.uuid
+          self.bulk_import_id = if klass.method_defined?(:generate_unique_id)
+                                  klass.generate_unique_id
+                                else
+                                  SecureRandom.uuid
+                                end
```
harsha-flipp, line 33: bulk_import_id currently has UUID. I want a way to override it as we find difference in performance with ULIDs instead of UUIDs.

There are two ways to do it.
- Defining method in klass (Active Record) associated to populate the bulk_import_id
- Passing the proc like key_cols (pushed relevant code changes as an example)


I prefer the first way and documenting that you can override it in relevant ActiveRecord class. The second way lets you override in Consumer instead. Since bulk_import_id is an active record attribute, I feel like doing it this way instead of passing procs around. What do you think Dan?

dorner: I'd prefer doing it in the consumer. I'd like to keep the model class as separate as possible from Deimos. Not sure why we did `key_cols` this way, but the other option is to define a method on the consumer `generate_bulk_import_id` that maybe can take a class or record as an argument.

dorner (changes requested):

lib/deimos/active_record_consume/batch_record.rb
harsha-flipp: `BatchConsumption` module uses `BatchRecord` and `MassUpdater` modules. I tried taking a pass at defining a method in consumer. To pass methods to those independent modules, creating a proc and passing it in constructor is the way. That is why, I believe, `key_cols` is passed as a proc in constructor.

https://github.com/flipp-oss/deimos/blob/92437aac91fc529348a5d840275138a1b377a50d/lib/deimos/active_record_consume/batch_consumption.rb#L190-L192
https://github.com/flipp-oss/deimos/blob/92437aac91fc529348a5d840275138a1b377a50d/lib/deimos/active_record_consume/batch_consumption.rb#L165-L168

What I could think of is.. We need to fill bulk_import_id in two cases.
1) To primary record when associations are present.
2) To sub records when replace_associations is required.

Case 1: I can fill bulk_import_id for primary records in BatchConsumption by moving that logic out of [BatchRecords](https://github.com/flipp-oss/deimos/blob/92437aac91fc529348a5d840275138a1b377a50d/lib/deimos/active_record_consume/batch_record.rb#L28-L30) into BatchConsumption module.
 
Case 2: I will still have to pass it to MassUpdater as it [updates data for sub-records](https://github.com/flipp-oss/deimos/blob/92437aac91fc529348a5d840275138a1b377a50d/lib/deimos/active_record_consume/mass_updater.rb#L72). Any ideas on how to refactor for this case?

dorner: You need to pass a proc, but you can **define** it as a method. That's actually what happens with `key_cols` - it's defined as a method and then turned into a proc:

```
        key_col_proc = self.method(:key_columns).to_proc
```

So you can define the method on the consumer class, but then pass it into the other classes as a proc.

harsha-flipp:
Lionel tackled with these in his other PRs. Closing this!

PR #201, opened 2023-08-24
post_process method outside batch consumption DB transaction

harsha-flipp:
# Pull Request Template

## Description

After batch consumption transaction is complete(BULK INSERTs), previously nothing is returned. Now we are passing the active record objects created back to consume_batch method so these records objects can be used by `post_process` hook method. Consumers can override this to do an action on the Active Record Models.

Note that these active record models are used by ActiveRecord-Import gem to bulk insert data. So record IDs are populated only if there is association involved.

Usecase: After saving records to Database, I queue a sidekiq job with record IDs.
Fixes # (issue)

## Type of change

## How Has This Been Tested?

I ran the unit tests locally and this code change adds an empty `post_process` method in consumer. So it should not hurt any one upgrading to this version. For a meaningful use, they will have to override `post_process` in their consumers.

- [x] Added an RSpec to note that the records are passed to post_process method
- [ ] Couldnot add an RSpec to try both the branches of compacted and uncompacted messages. But I tried it locally by setting the parameter on consumer class.

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -352,6 +352,11 @@ class MyBatchConsumer < Deimos::Consumer
       # Do something 
     end
   end
+
+  def post_process(records)
+    # ActiveRecord objects that have been used to save to the database.
+    # They contain primary keys only if associations are involved.
+  end
 end
 ```
 #### Saving data to Multiple Database tables
--- a/lib/deimos/active_record_consume/batch_consumption.rb
+++ b/lib/deimos/active_record_consume/batch_consumption.rb
@@ -22,7 +22,7 @@ module Deimos
       # they are split
       # @param payloads [Array<Hash,Deimos::SchemaClass::Record>] Decoded payloads
       # @param metadata [Hash] Information about batch, including keys.
-      # @return [void]
+      # @return [Array<ActiveRecord::Base>]
       def consume_batch(payloads, metadata)
         messages = payloads.
           zip(metadata[:keys]).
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -352,6 +352,11 @@ class MyBatchConsumer < Deimos::Consumer
       # Do something 
     end
   end
+
+  def post_process(records)
+    # ActiveRecord objects that have been used to save to the database.
+    # They contain primary keys only if associations are involved.
```
dorner, line 358: The wording is a bit awkward... maybe

```
This method gets passed the ActiveRecord objects that were saved to the database.
They contain primary keys only if associations were also saved via bulk_import_id.
```

lib/deimos/active_record_consume/batch_consumption.rb
```diff
@@ -114,45 +123,48 @@ def compact_messages(batch)
       # All messages are split into slices containing only unique keys, and
       # each slice is handles as its own batch.
       # @param messages [Array<Message>] List of messages.
-      # @return [void]
+      # @return [Array<ActiveRecord::Base>]
       def uncompacted_update(messages)
         BatchSlicer.
           slice(messages).
-          each(&method(:update_database))
+          each(&method(:update_database)).flatten
```
dorner, line 130: Probably should use `map` instead of `each` here since we're using the return value.

spec/active_record_consume/batch_consumption_spec.rb
```diff
@@ -100,4 +99,24 @@
...
+      let(:metadata) do
+        { keys: [1, 2, 3, 4, 5] }
+      end
+
+      it 'should be called once with a flat array of records' do
+        records = [Object, Object] # should be ActiveRecord object
+        expect(batch_consumer).to receive(:post_process).once.with(records)
+        expect(batch_consumer).to receive(:update_database) { records }
+        batch_consumer.send(:consume_batch, payload, metadata)
+      end
```
dorner, line 119: I think we're probably missing some more tests - e.g. with and without max\_db\_batch\_size, and with and without deletions. Probably can just add another `expect` onto existing tests showing that they return the expected values.

dorner (changes requested):

README.md
harsha-flipp: `have been used to` I am sorry I should not flex my grammar. I will use your words.

I wanted to give the intent that these are not the active records returned after save. These are the records we created and passing to save the record.

dorner: But... aren't they the same thing?

harsha-flipp: I misunderstood that rails adds IDs back to the objects for every save.

I tested in rails console and attribute ID is saved to the object only when`object.save!` is called. For other scenarios, `object.save` doesnot really change the object attribute ID.

harsha-flipp:
Lionel tackled with these changes in his PR. I am closing this.

PR #204, opened 2023-11-27
CCOL-2039: Support for post processing and custom batch generation

lionelpereira:
# Pull Request Template

## Description

Include current monkey patches currently implemented in Item Platform

1. Allow global and individual consumer batch id generator procs to be passed in for use in bulk imports
2. Allow global and individual consumer replace_associations configuration for use in bulk imports
3. Allow record lookup entity for use in record attribute construction.
Ex. Reference DB query capturing entire payload batch between record construction instead of individual queries when each record attributes are created
4. Return valid and invalid records saved during consumption for further processing

Fixes # (issue)

## Type of change

## How Has This Been Tested?

- [ ] Test A
- [ ] Test B

## Checklist:

### DIFF
--- a/lib/deimos/active_record_consume/batch_consumption.rb
+++ b/lib/deimos/active_record_consume/batch_consumption.rb
@@ -165,7 +165,8 @@ module Deimos
         updater = MassUpdater.new(@klass,
                                   key_col_proc: key_col_proc,
                                   col_proc: col_proc,
-                                  replace_associations: self.class.config[:replace_associations])
+                                  replace_associations: self.class.config[:replace_associations],
+                                  batch_id_generator: self.class.bulk_import_id_generator)
         updater.mass_update(record_list)
       end
 
--- a/lib/deimos/active_record_consume/batch_consumption.rb
+++ b/lib/deimos/active_record_consume/batch_consumption.rb
@@ -165,7 +165,7 @@ module Deimos
         updater = MassUpdater.new(@klass,
                                   key_col_proc: key_col_proc,
                                   col_proc: col_proc,
-                                  replace_associations: self.class.config[:replace_associations],
+                                  replace_associations: self.class.replace_associations,
                                   batch_id_generator: self.class.bulk_import_id_generator)
         updater.mass_update(record_list)
       end
--- a/lib/deimos/active_record_consume/batch_consumption.rb
+++ b/lib/deimos/active_record_consume/batch_consumption.rb
@@ -170,15 +170,27 @@ module Deimos
         updater.mass_update(record_list)
       end
 
+      # Returns a lookup entity to be used during record attributes
+      # @param _messages [Array<Deimos::Message>]
+      # @return [Hash, Set, ActiveRecord::Relation]
+      def record_lookup(_messages)
+        {}
+      end
+
       # @param messages [Array<Deimos::Message>]
       # @return [BatchRecordList]
       def build_records(messages)
+        lookup = record_lookup(messages)
         records = messages.map do |m|
-          attrs = if self.method(:record_attributes).parameters.size == 2
+          attrs = case self.method(:record_attributes).parameters.size
+                  when 3
+                    record_attributes(m.payload, m.key, lookup)
+                  when 2
                     record_attributes(m.payload, m.key)
                   else
                     record_attributes(m.payload)
                   end
+
           next nil if attrs.nil?
 
           attrs = attrs.merge(record_key(m.key))
--- a/lib/deimos/active_record_consume/batch_consumption.rb
+++ b/lib/deimos/active_record_consume/batch_consumption.rb
@@ -31,16 +31,19 @@ module Deimos
         tags = %W(topic:#{metadata[:topic]})
 
         Deimos.instrument('ar_consumer.consume_batch', tags) do
+          valid_upserts = []
+          invalid_upserts = []
           # The entire batch should be treated as one transaction so that if
           # any message fails, the whole thing is rolled back or retried
           # if there is deadlock
           Deimos::Utils::DeadlockRetry.wrap(tags) do
-            if @compacted || self.class.config[:no_keys]
-              update_database(compact_messages(messages))
-            else
-              uncompacted_update(messages)
-            end
+            valid_upserts, invalid_upserts = if @compacted || self.class.config[:no_keys]
+                                               update_database(compact_messages(messages))
+                                             else
+                                               uncompacted_update(messages)
+                                             end
           end
+          post_process(valid_upserts, invalid_upserts)
         end
       end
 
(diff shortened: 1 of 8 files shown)


lib/deimos/active_record_consume/batch_record.rb
```diff
@@ -22,11 +22,12 @@ class BatchRecord
       # @param klass [Class < ActiveRecord::Base]
       # @param attributes [Hash] the full attribute list, including associations.
       # @param bulk_import_column [String]
-      def initialize(klass:, attributes:, bulk_import_column: nil)
+      # @param bulk_id_generator [Proc]
+      def initialize(klass:, attributes:, bulk_import_column: nil, bulk_id_generator: nil)
```
dorner, line 26: Why is this sometimes called `bulk_id_generator` and sometimes `batch_id_generator`?

lib/deimos/config/configuration.rb
```diff
@@ -90,7 +90,9 @@ def self.configure_producer_or_consumer(kafka_config)
       if kafka_config.respond_to?(:bulk_import_id_column) # consumer
         klass.config.merge!(
           bulk_import_id_column: kafka_config.bulk_import_id_column,
-          replace_associations: kafka_config.replace_associations
+          replace_associations: kafka_config.replace_associations,
+          bulk_import_id_generator: kafka_config.bulk_import_id_generator ||
+            Deimos.config.consumers.bulk_import_id_generator
```
dorner, line 97: Do we have a use case for a global setting as well as a per-consumer setting? I feel like this would have to be something specified per consumer.

spec/active_record_consume/mass_updater_spec.rb
```diff
@@ -39,6 +39,8 @@
       end
     end
 
+    let(:bulk_id_generator) { proc { SecureRandom.uuid } }
```
dorner, line 42: Is this actually tested in the spec? This is identical to the default so how can we validate that it's actually being called?

lib/deimos/config/configuration.rb
```diff
@@ -90,7 +90,9 @@ def self.configure_producer_or_consumer(kafka_config)
       if kafka_config.respond_to?(:bulk_import_id_column) # consumer
         klass.config.merge!(
           bulk_import_id_column: kafka_config.bulk_import_id_column,
-          replace_associations: kafka_config.replace_associations,
+          replace_associations: kafka_config.replace_associations.nil? ?
+                                  Deimos.config.consumers.replace_associations :
+                                  kafka_config.replace_associations,
```
dorner, line 95: Is this purely so we don't have to specify this for each individual consumer?

lib/deimos/active_record_consume/batch_consumption.rb
```diff
@@ -170,15 +170,27 @@ def upsert_records(messages)
         updater.mass_update(record_list)
       end
 
+      # Returns a lookup entity to be used during record attributes
+      # @param _messages [Array<Deimos::Message>]
+      # @return [Hash, Set, ActiveRecord::Relation]
+      def record_lookup(_messages)
```
dorner, line 194: I don't really understand what this is for. Can we explain this in the README?

I'd rather not make changes to the `record_attributes` method if possible. I'm not sure what this is, but can it be put in metadata somehow? Or maybe give `Deimos::Message` a `metadata` field for custom data and have a pre-process method to populate it?

lib/deimos/active_record_consume/batch_consumption.rb
```diff
@@ -132,30 +135,39 @@ def update_database(messages)
...
+                                               invalid_upserts.push(invalid)
+                                             end
+                                             valid_upserts.compact!
+                                             invalid_upserts.compact!
+                                           else
+                                             upsert_records(upserted)
+                                           end
+        end
+
+        unless removed.empty?
```
dorner, line 154: `if removed.any?`

lib/deimos/active_record_consume/batch_consumption.rb
```diff
@@ -132,30 +135,39 @@ def update_database(messages)
...
 
       # Upsert any non-deleted records
       # @param messages [Array<Message>] List of messages for a group of
       # records to either be updated or inserted.
-      # @return [void]
+      # @return [Array<Array<ActiveRecord::Base>, Array<BatchRecord>]
       def upsert_records(messages)
         record_list = build_records(messages)
-        record_list.filter!(self.method(:should_consume?).to_proc)
+        invalid = filter_records(record_list)
```
dorner, line 170: This doesn't seem correct. `filter_records` just returns the mutated list of records where we've already deleted things that `should_consume?` returns false. It doesn't return any list of invalid records. Unless I'm reading this wrong.

lib/deimos/active_record_consume/batch_record_list.rb
```diff
@@ -23,6 +23,24 @@ def filter!(method)
...
+          case method.parameters.size
+          when 2
+            self.batch_records.select! do |record|
+              method.call(record.record, record.associations)
+            end
+          else
+            self.batch_records.select! { |record| method.call(record.record) }
+          end
+        end
+      end
```
dorner, line 42: Not sure I understand why we need this complex method? Also it's super confusing - in regular Ruby, `delete_if` and `reject!` are almost identical. And `reject!` is the *oppposite* of `select!`, so... what's going on?

lib/deimos/active_record_consume/mass_updater.rb
```diff
@@ -83,9 +83,11 @@ def import_associations(record_list)
       end
 
       # @param record_list [BatchRecordList]
+      # @return [Array<ActiveRecord::Base>]
       def mass_update(record_list)
         save_records_to_database(record_list)
         import_associations(record_list) if record_list.associations.any?
+        record_list.records
```
dorner, line 90: Can a test be added for this?

lib/deimos/active_record_consumer.rb
```diff
@@ -75,8 +75,9 @@ def initialize
     # attributes set to the new/existing record.
     # @param payload [Hash,Deimos::SchemaClass::Record]
     # @param _key [String]
+    # @param _lookup [Set, ActiveRecord::Relation, Hash]
     # @return [Hash]
-    def record_attributes(payload, _key=nil)
+    def record_attributes(payload, _key=nil, _lookup=nil)
```
dorner, line 80: Can we add a test for this feature?

lib/deimos/config/configuration.rb
```diff
@@ -451,7 +458,7 @@ def self.configure_producer_or_consumer(kafka_config)
       setting :bulk_import_id_column, :bulk_import_id
       # If true, multi-table consumers will blow away associations rather than appending to them.
       # @return [Boolean]
-      setting :replace_associations, true
+      setting :replace_associations, nil
```
dorner, line 461: Can we add a test to validate this feature?

dorner (changes requested):

lib/deimos/active_record_consume/batch_record.rb
lionelpereira: Updated

lib/deimos/config/configuration.rb
lionelpereira: Just for convenience. All of of our consumers use the same batch generator (Ulids)

lib/deimos/config/configuration.rb
lionelpereira: Pretty much

lib/deimos/config/configuration.rb
lionelpereira: Added a spec

lib/deimos/active_record_consume/batch_consumption.rb
lionelpereira: In item platform, before consuming local offers, we preload and create a lookup relation where we map merchant_id and merchant_store_code to a store id. 
```

store_set = messages.map { |m|
        { merchant_id: m.payload.merchant_id, merchant_store_code: m.payload.store_code }
      }.to_set
stores= store_set.reduce(Store.none) { |store, h| store.or(Store.where(h)) }.
        select(:id, :merchant_store_code, :merchant_id)
```
In record_attributes
```
id = stores.
        find { |store|
          store.merchant_store_code == store_code &&
            store.merchant_id == merchant_id
        }&.id
```

I think it makes sense to do this lookup and assignment in a pre_process method and attach the result to the deimos message as you suggested.
I'll look into implementing that

lib/deimos/active_record_consume/batch_record_list.rb
lionelpereira: This should all be `reject`` 😞 No clue what brain farts led to this

lionelpereira: Changed this to use `partition` instead

lib/deimos/active_record_consumer.rb
lionelpereira: Removed lookup

lib/deimos/active_record_consume/mass_updater.rb
lionelpereira: Added a spec

PR #207, opened 2023-11-30
CCOL-2039: Post process valid and invalid records from batch consumption

lionelpereira:
# Pull Request Template

## Description

1. `should_consume?` method accepts validation of associations
2. Return valid and invalid records saved during consumption for further processing in post_process method

Fixes # (issue)

## Type of change

## How Has This Been Tested?

- [ ] Test A
- [ ] Test B

## Checklist:


lib/deimos/active_record_consume/batch_consumption.rb
```diff
@@ -114,50 +117,64 @@ def compact_messages(batch)
...
+        invalid_upserts = []
         if upserted.any?
+          valid_upserts, invalid_upserts = if max_db_batch_size
+                                             upserted.each_slice(max_db_batch_size) do |group|
+                                               valid, invalid = upsert_records(group)
+                                               valid_upserts.push(*valid)
+                                               invalid_upserts.push(*invalid)
+                                             end
+                                             valid_upserts.compact!
+                                             invalid_upserts.compact!
```
dorner, line 153: Is there a reason you're compacting inside each iteration instead of after all iterations are done?

lib/deimos/active_record_consume/batch_consumption.rb
```diff
@@ -166,7 +183,13 @@ def upsert_records(messages)
                                   key_col_proc: key_col_proc,
                                   col_proc: col_proc,
                                   replace_associations: self.class.config[:replace_associations])
-        updater.mass_update(record_list)
+        [updater.mass_update(record_list), invalid]
+      end
+
+      # @param record_list [BatchRecordList]
+      # @return [Array<BatchRecord>]
```
dorner, line 190: Isn't this an array of arrays?

lib/deimos/active_record_consume/batch_record_list.rb
```diff
@@ -23,6 +23,29 @@ def filter!(method)
         self.batch_records.delete_if { |record| !method.call(record.record) }
       end
 
+      # Filter and return removed invalid batch records by the specified method or block
+      # @param method [Proc]
+      # @param block [Proc]
+      # @return [Array<BatchRecord>]
+      def partition!(method=nil, &block)
+        valid, invalid = if method.nil?
```
dorner, line 31: This should never be `nil` since it's defined on the base class, no?

spec/active_record_batch_consumer_spec.rb
```diff
@@ -493,5 +493,144 @@ def record_attributes(payload, key)
...
+          expect(widget_one.some_int).to eq(1)
+          expect(widget_two.some_int).to eq(20)
+          expect(widget_three.some_int).to eq(2000)
+          expect(widget_three.test_id).to eq(widget_two.test_id)
+          expect(widget_four.some_int).to eq(11)
+          expect(widget_four.test_id).to eq(widget_one.test_id)
+        end
+      end
+
+      context 'with post processing' do
```
dorner, line 546: I'm confused - what's the difference between `describe 'post processing'` and `context 'with post processing'`?

lib/deimos/active_record_consume/batch_record_list.rb
dorner: Not sure... I think I'd need to see this. It feels like you're diving too deep into Deimos guts to get this working - it feels wrong for calling code to have to call so many internal methods.

dorner (changes requested):
I still don't fully understand the use case here. We seem to be adding a lot of functionality for a very laser-specific use case.

lib/deimos/active_record_consume/batch_consumption.rb
```diff
@@ -166,7 +170,18 @@ def upsert_records(messages)
...
+      def filter_records(record_list)
+        record_list.filter!(self.method(:should_consume?).to_proc, consume_filter)
+      end
+
+      # @return [ActiveRecord::Relation,Hash,Set]
+      def consume_filter
```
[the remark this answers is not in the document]
lionelpereira: Creates a lookup entity for use when filtering records. 
Example:
```
def consume_filter
   Widget.where(data_source: 'datafeed')
end

def should_consume?(record, filter)
   record.valid?  && filter.exlude?(record)
end
```

This is intended for use after the record is built instead of having this filtering occur in `pre_process` instead

lib/deimos/active_record_consume/batch_consumption.rb
```diff
@@ -41,9 +42,17 @@ def consume_batch(payloads, metadata)
               uncompacted_update(messages)
             end
           end
+          process_valid_records(@valid_active_records)
```
dorner, line 45: Is there a reason this also can't be a notification?

lib/deimos/active_record_consume/batch_consumption.rb
```diff
@@ -92,9 +101,9 @@ def deleted_query(records)
         @klass.unscoped.where(@klass.primary_key => keys)
       end
 
-      # @param _record [ActiveRecord::Base]
+      # @param _batch_record [BatchRecord]
       # @return [Boolean]
-      def should_consume?(_record)
+      def should_consume?(_batch_record)
```
dorner, line 106: Why did this need to change? Did it actually change?

lib/deimos/active_record_consume/batch_consumption.rb
```diff
@@ -155,8 +164,14 @@ def update_database(messages)
       # @return [void]
       def upsert_records(messages)
         record_list = build_records(messages)
-        record_list.filter!(self.method(:should_consume?).to_proc)
+        invalid = filter_records(record_list)
+        if invalid.any?
+          ActiveSupport::Notifications.instrument('batch_consumption.invalid_records', {
```
dorner, line 169: So this doesn't actually move the processing out of the transaction. I'm honestly still a little confused about how important this is. If it's taking a really long time to log invalid messages, you probably shouldn't be doing it? Or we should be looking into why it takes so long?

lib/deimos/instrumentation.rb
```diff
@@ -80,4 +80,9 @@ def self.send_produce_error(event)
     event = ActiveSupport::Notifications::Event.new(*args)
     KafkaListener.send_produce_error(event)
   end
+
+  ActiveSupport::Notifications.subscribe('batch_consumption.invalid_records') do |*args|
+    payload = ActiveSupport::Notifications::Event.new(*args).payload
+    payload[:consumer].process_invalid_records(payload[:records])
```
dorner, line 86: Calling code should be able to do this themselves rather than us having a method that we call explicitly.

dorner (changes requested):

lib/deimos/active_record_consume/batch_consumption.rb
lionelpereira: This could also be a notification

lib/deimos/active_record_consume/batch_consumption.rb
lionelpereira: It did change types from `ActiveRecord::Base` to `BatchRecord`. This is in order to validate `associations` that are not yet saved

lib/deimos/active_record_consume/batch_consumption.rb
lionelpereira: Its important if we want to log datadog metrics for failed records. As of now its currently logging within the transaction but it would be nice to have this live outside the transaction if possible.

I can discuss with the team the current usefulness of the metrics outside of just logging the results

dorner (changes requested):

lib/deimos/active_record_consume/batch_consumption.rb
```diff
@@ -28,18 +28,13 @@ def consume_batch(payloads, metadata)
...
-          # The entire batch should be treated as one transaction so that if
-          # any message fails, the whole thing is rolled back or retried
-          # if there is deadlock
-          Deimos::Utils::DeadlockRetry.wrap(tags) do
-            if @compacted || self.class.config[:no_keys]
-              update_database(compact_messages(messages))
-            else
-              uncompacted_update(messages)
-            end
+        @tags = %W(topic:#{metadata[:topic]})
```
dorner, line 31: I think it's weird to pass metric tags so far down into the workflow. Can we do something like

```
Deimos.config.tracer.active_span.set_tag("topic", metadata[:topic])
...
Deimos.config.tracer.active_span.get_tag("topic")
```

We'd have to add a `get_tag` method to the `provider` class but it feels a lot better because we can separate out the tracing/metric logic from the actual business logic.

lib/deimos/active_record_consume/batch_consumption.rb
```diff
@@ -155,19 +151,35 @@ def update_database(messages)
       # @return [void]
       def upsert_records(messages)
         record_list = build_records(messages)
-        record_list.filter!(self.method(:should_consume?).to_proc)
+        invalid = filter_records(record_list)
+        unless invalid.blank?
```
dorner, line 155: Better to use positive conditions: `if invalid.any?`

dorner (changes requested):

spec/active_record_batch_consumer_spec.rb
```diff
@@ -255,64 +255,6 @@ def deleted_query(_records)
       end
     end
 
-    describe 'batch atomicity' do
```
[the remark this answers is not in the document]
lionelpereira: Wouldn't this spec be handled by the deadlock retry spec

lib/deimos/active_record_consume/batch_consumption.rb
```diff
@@ -28,15 +28,16 @@ def consume_batch(payloads, metadata)
           zip(metadata[:keys]).
           map { |p, k| Deimos::Message.new(p, nil, key: k) }
 
-        @tags = %W(topic:#{metadata[:topic]})
+        Deimos.config.tracer.active_span.set_tag('topic', metadata[:topic])
 
-        Deimos.instrument('ar_consumer.consume_batch', @tags) do
+        Deimos.instrument('ar_consumer.consume_batch',
+                          Deimos.config.tracer.active_span.get_tag('topic')) do
```
dorner, line 34: lol you don't need to do `get_tag` here, you already have the topic.

CHANGELOG.md
```diff
@@ -12,6 +12,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Feature: Add individual configuration for custom `bulk_import_id_generator` proc per consumer
 - Feature: Add global `replace_assocations` value for  for all consumers
 - Feature: Add individual `replace_assocations` value for for individual consumers
+- Feature: `should_consume?` method accepts BatchRecord associations
+- Feature: Reintroduce `filter_records` for bulk filtering of records prior to insertion
+- Feature: Return valid and invalid records saved during consumption for further processing in `batch_consumption.valid_records` and `batch_consumption.invalid_records` ActiveSupport Notifications
```
dorner, line 17: This should be added to the README.

lib/deimos/tracing/provider.rb
```diff
@@ -42,6 +42,12 @@ def set_tag(tag, value, span=nil)
         raise NotImplementedError
       end
 
+      # Get a tag from a span with the specified tag.
+      # @param tag [String]
+      def get_tag(tag)
+        raise NotImplementedError
+      end
```
dorner, line 49: We need to implement this for datadog.

dorner (changes requested):

dorner (dismissed):

lib/deimos/consumer.rb
```diff
@@ -19,8 +19,17 @@ class Consumer
...
-        @decoder ||= Deimos.schema_backend(schema: config[:schema],
-                                           namespace: config[:namespace])
+        return @decoder if @decoder
+
+        @decoder = if Utils::SchemaClass.use?(config.to_h)
+                     # Initialize an instance of the provided schema
+                     # in the event the schema class is an override, the inherited
+                     # schema and namespace will be applied
+                     schema_class = "Schemas::#{config[:schema]}".constantize.new
+                     Deimos.schema_backend(schema: schema_class.schema, namespace: schema_class.namespace)
```
dorner, line 29: Shouldn't this be moved into the `schema_backend` method rather than just used here?

lib/deimos/tracing/datadog.rb
```diff
@@ -11,15 +11,12 @@ def initialize(config)
         raise 'Tracing config must specify service_name' if config[:service_name].nil?
 
         @service = config[:service_name]
+        @tracer = ::Datadog.respond_to?(:tracer) ? ::Datadog.tracer : ::Datadog::Tracing
```
dorner, line 14: This feels more like a method than an instance variable. It isn't actually state that needs to be stored.

spec/consumer_spec.rb
```diff
@@ -21,6 +22,33 @@ def consume(_payload, _metadata)
...
+
+        def as_json(_opts={})
+          {
+            'test_id' => @test_id,
+            'some_int' => @some_int,
+            'payload_key' => @payload_key&.as_json
+          }
+        end
+      end
+      stub_const('Schemas::MySchema', schema_class)
```
dorner, line 51: Is there a reason we have to add this in all four specs? If we're just trying to test overriding, can we have a dedicated spec for that? Also, why does the overriding class have to reimplement every method?

dorner (changes requested):

spec/consumer_spec.rb
lionelpereira: `Schemas::MySchema` wasn't defined anywhere. So in tests where `SCHEMAS_CLASS_SETTING = on` test were failing.

dorner: It should be in `spec/schemas/my_namespace`. If it's not showing up, maybe the settings are wrong for the tests? Or do we need to add it to `spec/schemas` and require it similarly to the other generated schema files?

spec/schemas/my_namespace/my_schema_updated.rb
dorner, line 1: Filename is backwards :)

dorner (changes requested):
Looks much better! One tiny comment left :)

spec/schemas/my_namespace/my_schema_updated.rb
lionelpereira: Oof. Updated lol

dorner (approved):

dorner:
Time to hit the big green button! Thanks @lionelpereira !

PR #208, opened 2023-12-20
Feature/schema namespace map

newbray:
## Description

The current approach for namespacing schema classes (just use the last bit of the namespace) is not sufficient for certain naming schemes. For example:

- 'com.company.accounts.events' and 'com.company.bookings.events' would lump everything into the `events` module / directory.

It would be useful if we could use the entire namespace but be able to configure and simplify it as Avro schema namespaces are long and not very ruby-esque.

Changes in this MR:
- Consider entire namespace when generating schemas
- Implement schema_namespace_map to enable namespace configuration as per discussion here https://github.com/flipp-oss/deimos/issues/153#issuecomment-1093080328
- Some zeitwerk fixes for plural module names

Example usage:

```
schema do
  generated_class_path 'app/schema_classes'
  schema_namespace_map({
    'com.company.sub_org' => 'SchemaClasses', # anything matching com.company.sub_org can be squashed
    'com.company.sub_org.avro' => 'SchemaClasses', # I don't want an avro module thanks!
    'com.company.sub_org.avro.something' => 'SchemaClasses' # urgh whoever added this schema REALLY likes long namespaces
  })
  generate_namespace_folders true    
end
```

## Type of change

## How Has This Been Tested?

Added spec

## Checklist:


lib/deimos/utils/schema_class.rb
```diff
@@ -10,10 +10,23 @@ class << self
         # @return [Array<String>]
         def modules_for(namespace)
           modules = ['Schemas']
-          namespace_folder = namespace.split('.').last
-          if Deimos.config.schema.generate_namespace_folders && namespace_folder
-            modules.push(namespace_folder.underscore.classify)
+
+          if Deimos.config.schema.generate_namespace_folders
+            namespace_map = Deimos.config.schema.schema_namespace_map.sort_by{|k,v| -k.length}.to_h
+            namespace_override = namespace_map.keys.detect{ |k| namespace.include?(k) }
```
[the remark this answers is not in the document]
dorner: Also - this indicates that the namespace map is actually a map of namespace **prefixes** rather than fully articulated namespaces. If that's what's intended, we should make sure the documentation is clear about this. Or even rename this to `namespace_prefix_map`.

lib/deimos/utils/schema_class.rb
```diff
@@ -10,10 +10,23 @@ class << self
...
+
+            module_namespace = if namespace_override.present?
+              # don't use Schemas default module
```
dorner, line 19: This also should be called out in documentation. Is there a reason we don't want to use the default module as the base?

lib/deimos/utils/schema_class.rb
```diff
@@ -10,10 +10,23 @@ class << self
...
+              modules = Array(namespace_map[namespace_override].dup)
```
dorner, line 20:
```suggestion
              modules = [namespace_map[namespace_override]]
```

lib/deimos/utils/schema_class.rb
```diff
@@ -10,10 +10,23 @@ class << self
...
           modules = ['Schemas']
-          namespace_folder = namespace.split('.').last
-          if Deimos.config.schema.generate_namespace_folders && namespace_folder
-            modules.push(namespace_folder.underscore.classify)
+
+          if Deimos.config.schema.generate_namespace_folders
+            namespace_map = Deimos.config.schema.schema_namespace_map.sort_by{|k,v| -k.length}.to_h
+            namespace_override = namespace_map.keys.detect{ |k| namespace.include?(k) }
+
+            module_namespace = if namespace_override.present?
```
dorner, line 18: This block is setting a return value and also doing side effects... I'd prefer this in the format of

```
module_namespace = namespace
if namespace_override.present?
  module_namespace = namespace.gsub(/#{namespace_override}\.?/, '')
  modules = [namespace_map[namespace_override]]
end
```

lib/deimos/utils/schema_class.rb
```diff
@@ -10,10 +10,23 @@ class << self
...
+              # don't use Schemas default module
+              modules = Array(namespace_map[namespace_override].dup)
+              namespace.gsub(/#{namespace_override}.?/, '')
```
dorner, line 21: Shouldn't this be `\.` instead of `.` in the regex? We're specifically replacing the `.` character, no?

lib/deimos/utils/schema_class.rb
```diff
@@ -10,10 +10,23 @@ class << self
...
+            else
+              namespace
+            end
+
+            namespace_folders = module_namespace.split('.').map{ |f| f.underscore.camelize }
```
dorner, line 26: Why the move from `classify` to `camelize`? This is a breaking change.

dorner (changes requested):
I like the overall approach! Had some comments about the implementation.

newbray:
> I like the overall approach! Had some comments about the implementation.

Hey, thanks for reviewing this so quickly! I'll address the comments in the new year. Hope you have a great Christmas! 🎄

lib/deimos/utils/schema_class.rb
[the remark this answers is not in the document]
newbray: good point thanks, will update the docs

lib/deimos/utils/schema_class.rb
newbray: Will update the docs. The reason to not use the default is you may want to use a different module name so it makes it more configurable.

For example, we have the Avro schemas in a schemas directory and the generated classes in a schema_classes directory so we want our namespace map to look like this:
```
schema_namespace_map({
        'com.org.suborg' => 'SchemaClasses',
        'com.org.suborg.avro' => 'SchemaClasses', # no need to include an avro module
        'com.org.suborg.avro.somethingelse' => 'SchemaClasses' # too much nesting otherwise
      })
```
For our use case, we really just need a configurable array of namespace prefixes to squash and the ability to set the default module, but having the map makes it highly configurable for other use cases.

lib/deimos/utils/schema_class.rb
newbray: I want the array constructor so that you don't have to provide an array in the config when you just want a single module ('com.org.suborg' => 'SchemaClasses') and so that it is also compatible with multiple modules ('com.org.suborg' => ['SchemaClasses', 'SubOrg']).

The `dup` was necessary but I removed it in [c80da4b](https://github.com/flipp-oss/deimos/pull/208/commits/c80da4b4907ba49ef921888a8a497e8e12e46853) as this change meant it was no longer needed

lib/deimos/utils/schema_class.rb
newbray: sure, will update

lib/deimos/utils/schema_class.rb
newbray: This is because module names can be plural. If you have a bunch of log related schemas under the namespace `com.my_org.logs` you would want them in the Logs module rather than Log.

I see where you're coming from with this being a breaking change, however simply updating the gem will not break anything.

When you regenerate the classes, if you previously used the `generate_namespace_folders` option you are going to have to modify references anyway as this PR changes that to use the entire namespace.

The modules would change from the current `Schemas => Log` to `Schemas => Com => MyOrg => Logs`. You could alternatively use the new config option to keep the previous behaviour like this:

```
schema_namespace_map({
        'com.my_org.logs' => ['Schemas', 'Log']
      })
```

dorner: Not sure I understand why the default behavior changed? (I must have missed that in the original PR.) If the point of this PR is to allow overriding the current behavior, why would we _also_ change the default?

lib/deimos/config/configuration.rb
```diff
@@ -381,6 +381,12 @@ def self.configure_producer_or_consumer(kafka_config)
       # Set to true to generate folders matching the last part of the schema namespace.
       # @return [Boolean]
       setting :generate_namespace_folders, false
+
+      # Use this option to reduce nesting when using generate_namespace_folders.
+      # For example: { 'com.mycompany.suborg' => 'SchemaClasses' }
+      # would replace a namespace matching the given key with the module name SchemaClasses.
```
dorner, line 387: namespace _prefixed with_ the given key

lib/generators/deimos/schema_class_generator.rb
```diff
@@ -122,7 +122,13 @@ def write_file(schema, key_config)
 
           file_prefix = schema.name.underscore.singularize
           if Deimos.config.schema.generate_namespace_folders
-            file_prefix = "#{@modules.last.underscore.singularize}/#{file_prefix}"
+            # Use entire namespace for folders
+            # but don't add directories that are already in the path
+            directories = @modules.map { |m|
+              Deimos.config.schema.generated_class_path.include?(m.underscore) ? nil : m.underscore
+            }.compact
```
dorner, line 129: Wouldn't this be clearer as
`directories = @modules.map(&:underscore).select { |m| Deimos.config.schema.generated_class_path.exclude?(m) }`

dorner (changes requested):
Couple more comments/questions. For next time, can you please avoid force pushing the PR? It makes it harder to see what changed since the last review. We can always squash it later. Thanks!

lib/deimos/utils/schema_class.rb
newbray: well currently it only uses the very last part of the namespace for modularization. For this configuration to be useful we need to use the entire namespace as one of the issues is having similarly named schemas conflicting. To give a made up example:
```
com.my-org.client.logs
com.my-org.customer.logs
com.my-other-org.customer.logs
```

All of these would currently be under Schemas => Log and would conflict. If we don't take into account the entire namespace there is no way to differentiate them or provide ways to override them.

One of the things I considered when deciding where we want our schema classes to be generated is to put them alongside the avro definitions, which must be fully namespaced. This change also would enable users of the gem to do that should they wish.

IMO this is the most flexible approach that allows users to have their generated files modularized in the way that suits their codebase

lib/generators/deimos/schema_class_generator.rb
newbray: yes thx, will update

lib/deimos/utils/schema_class.rb
newbray: I thought about this some more and I think we would be able to keep the current behaviour and it probably makes sense to do so. I'll see if I can get a commit together and we can see how that looks

newbray: ok, so I have some code that will accomplish this. The caveat here is that the namespace map config will need to be updated more frequently as a result. I'll try to explain what I mean with an example:

Let's say we have the following map:
```
schema_namespace_map({
    'com.company.sub_org' => 'SchemaClasses',
    'com.company.sub_org.timekeeping' => ['SchemaClasses', 'Timekeeping']
    'com.company.sub_org.hiring' => ['SchemaClasses', 'Hiring']
  })
```

And we have schemas `com.company.sub_org.timekeeping.some-schema` and `com.company.sub_org.hiring.some-other-schema`

But now we want to add a schema for the payroll domain `com.company.sub_org.payroll.some-schema`. We're going to need to add this to the config otherwise it will be generated in the SchemaClasses directory which is not ideal.

In fact every time we add a new schema with a different sub namespace we will need to also change the config. This was what I was trying to avoid by changing the existing behaviour.

My personal preference would be to leave this as it is by changing the default to use the full namespace as it seems more of a convention over configuration approach. That said, I'm happy to change it if necessary and will wait to hear what you think

dorner:
Hi - not ignoring you, just trying to find the time and mental energy to review this :smile:

lib/deimos/utils/schema_class.rb
dorner: Sorry for the delay. 

I understand the desire here. I'm wondering if we can do something like... if `schema_namespace_map` is used at all, then change the default to what you're describing. But that seems like it's too much magic. 

Could we add another config for this? I know this ends up having lots of different options around how/where schemas are generated.

Actually... could we repurpose `generate_namespace_folders` to control this instead? From what I understand you're the only folks using this option :grin: 

I think we have an option when we do a version bump to change the default behavior and possibly deprecate/remove the config. I'm thinking about doing this when we (ultimately) move from ruby-kafka/phobos to ruby-rdkafka as ruby-kafka is deprecated and unmaintained. Maybe you can add a comment to that effect.

newbray: yes I think renaming generate_namespace_folders to use_full_namespace is the best option here - I'll add a commit to do that

newbray:
hey @dorner any chance you could check over this again pls?

dorner (approved):
I think this is good to go! We'll do a minor release bump so we can make the breaking change clear.

newbray:
> I think this is good to go! We'll do a minor release bump so we can make the breaking change clear.

thanks :)

PR #212, opened 2024-03-26
Fix: ActiveRecordConsumer batch consumption was not working with the …

dorner:
…trilogy adapter.

### DIFF
--- a/deimos-ruby.gemspec
+++ b/deimos-ruby.gemspec
@@ -25,7 +25,7 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency('sigurd', '>= 0.1.0', '< 1.0')
 
   spec.add_development_dependency('activerecord-import')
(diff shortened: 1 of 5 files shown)


skawaguchi-flipp (approved):
LGTM, but I'm not sure how to test to make sure it's good.

deimos-ruby.gemspec
```diff
@@ -25,6 +25,7 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency('sigurd', '>= 0.1.0', '< 1.0')
 
   spec.add_development_dependency('activerecord-import')
+  spec.add_development_dependency('activerecord-trilogy-adapter', '> 0')
```
skawaguchi-flipp, line 28: question: How come `> 0`? I haven't seen this before.

dorner: Probably don't need it, it's the same as the above line that just has no condition.

Issue #219, opened 2024-05-31
2.0 Migration - Phobos -> Karafka


dorner:
This is a placeholder for the work needed for the v2 branch, which is a rewrite of Deimos backed by Karafka instead of Phobos. This is a large-scale rewrite.

@newbray @2rba @yannp @mikdiet @iMacTia - if you're still using this library, please take a look at the [v2 upgrade guide](https://github.com/flipp-oss/deimos/blob/v2/docs/UPGRADING.md) and let me know if there are any breaking changes that are blockers for you. In particular, there are a few features I removed because I knew we weren't using them in our company - I can definitely look at bringing them back if they are used elsewhere.

Please feel free to comment on this issue if you have questions or anything else you'd like to discuss!

iMacTia:
This is great news! Congrats on the huge milestone 👏

We're still very much using Deimos so I'll definitely take a look next week and let you know if there's anything critical we rely on that is going away ❤️

iMacTia:
@dorner I read the v2 upgrade guide and couldn't really see anything there that would cause us issues.
Once v2 is ready for production usage drop us a message and we'll be happy to give it a go

dorner:
Will do! This is currently blocked on https://github.com/karafka/karafka/pull/2108 as well as internal review and testing. I'll update this issue with info as it changes. Thanks!

newbray:
amazing, thanks 🙌

PR #220, opened 2024-05-31
V2 rebased

dorner:
Work in progress for 2.0 version of Deimos backed by Karafka. For more info, see https://github.com/flipp-oss/deimos/issues/219 .


spec/batch_consumer_spec.rb
```diff
@@ -251,18 +251,11 @@ def consume_batch(_payloads, _metadata)
...
-                   { key: 1, message_id: 'one' },
-                   { key: 2, message_id: 'two' }
-                 ]
-               )).
-          twice
+        allow(Deimos::Logging).to receive(:log_info)
+
+        expect(Deimos::Logging).
+          to receive(:log_info).
+          with(hash_including(payload_keys: ["1", "2"]))
```
harsha-flipp, line 215: `it 'should log message identifiers'`. The expectation does not seem to test the `message_id` part and removes it.

I think the title should change if you say that this is the correct way in v2.

.rubocop_todo.yml
```diff
@@ -159,12 +158,6 @@ Metrics/AbcSize:
     - 'lib/deimos/utils/schema_controller_mixin.rb'
```
harsha-flipp, line 158: inline_consumer and schema_controller_mixin can be removed as the files are removed in the PR.
```
    - 'lib/deimos/utils/inline_consumer.rb'
    - 'lib/deimos/utils/schema_controller_mixin.rb'
```

lib/deimos.rb
```diff
@@ -112,6 +115,31 @@ def start_db_backend!(thread_count: 1)
...
+        karafka_configs.find { |t| t.name == topic}
+      elsif producer
+        karafka_configs.find { |t| t.producer_class == producer}
+      end
+    end
+
+    # @param handler_class [Class]
+    # @return [String,nil]
+    def topic_for_consumer(handler_class)
+      Deimos.karafka_configs.each do |topic|
```
harsha-flipp, line 171: Why is this not `karafka_configs.each` like in the above method? They both should be referring to the same method, isn't it?

```
    def karafka_config_for(topic: nil, producer: nil)
      if topic
        karafka_configs.find { |t| t.name == topic}
      elsif producer
        karafka_configs.find { |t| t.producer_class == producer}
      end
    end
```

PR #229, opened 2025-05-22
Fix: Allow for hot reloading of producer classes without crashing.

dorner:


harsha-flipp (approved):
Thank you! We should release this.

PR #230, opened 2025-05-26
Fix: Outbox producer would crash if no Deimos producer was defined

dorner:


CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## UNRELEASED
 
+- Fix: Outbox producer would crash if no Deimos producer was defined for the topic.
```
harsha-flipp, line 10: What is this scenario where no producer will be defined for the topic. Does outbox not require me to define "producer block" in Karafka.rb?

dorner: In this case, the messages are being inserted by the old system but published by the new system (which doesn't have configurations for them).

harsha-flipp (approved):

harsha-flipp (approved):

PR #235, opened 2025-07-17
Make schema classes backwards-compatible

dorner:
Avro compatibility allows for adding fields with a default. When this happens, schema classes will crash because it doesn't recognize the field in the `initialize` method.

This change updates how the consumer specifically instantiates these messages so it simply ignores any fields it doesn't know about.

### DIFF
--- a/lib/deimos/schema_class/record.rb
+++ b/lib/deimos/schema_class/record.rb
@@ -72,11 +72,30 @@ module Deimos
         validator.schema_fields.map(&:name)
       end
 
+      # Used internally within Deimos so that we don't crash on unknown fields that come from
+      # a backwards compatible schema.
+      # @param kwargs [Hash] the attributes to set on the new object.
       # @return [SchemaClass::Record]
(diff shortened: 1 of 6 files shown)


lib/deimos/schema_class/record.rb
```diff
@@ -72,11 +72,30 @@ def schema_fields
...
+      end
+
+      # @return [SchemaClass::Record]
+      # @param from_message [Boolean] whether it's being initialized from a real Avro message.
+      def self.initialize_from_value(value, from_message: false)
         return nil if value.nil?
 
-        value.is_a?(self) ? value : self.new(**value.symbolize_keys)
+        return value if value.is_a?(self)
+        if from_message
```
nhaidar, line 94: is this backwards? If its from an avro message it should use `.new_from_message` not `.new` right?

dorner: You are correct - good catch!

nhaidar (approved):

PR #256, opened 2026-04-17
Fix: import_associations reuses stale parent id on DeadlockRetry

josephyang0519:
## Description

In the non-`save_associations_first` path of `MassUpdater#mass_update`, a deadlock on a child import (not on the parent) leaves the in-memory `record[primary_col]` populated after the transaction rolls back. That value was written by `fill_primary_keys!` during attempt 1. The existing guard `record[primary_col].present?` then short-circuits `fill_primary_keys!` on retry, and child records are built with the stale parent id.

In DBs that burn auto-increment on rollback (MySQL InnoDB, Postgres), the retry's parent INSERT lands on a different id and the child INSERT raises `ActiveRecord::InvalidForeignKey`.

```
Attempt 1
BEGIN TRANSACTION
  Step 1: INSERT parent → id = 202 in the DB
  Step 2: fill_primary_keys! → Ruby object now holds "parent.id = 202"
  Step 3: INSERT details → ok
  Step 4: INSERT distributions → DEADLOCK
ROLLBACK
After the rollback:
- In the DB: parent row 202 is gone (it was rolled back)
- In Ruby memory: the parent object still says "parent.id = 202"
- In MySQL: the counter that assigns IDs does not rewind. Next INSERT will get 203, not 202.
Attempt 2 (Deimos automatically retries after a deadlock)
BEGIN TRANSACTION
  Step 1: INSERT parent → id = 203 in the DB (because 202 was used up)
  Step 2: fill_primary_keys! → SKIPPED (Deimos checks "is parent.id already set in memory?" → yes, 202 → skips the lookup)
  Step 3: INSERT details with generic_offer_id = 202 → FOREIGN KEY FAILS (no row 202 exists, real parent is at 203)
ROLLBACK
```
We saw this on item-platform's print offer consumer - [Datadog trace](https://app.datadoghq.com/dashboard/py5-397-mkp?query=service%3Aitem_platform%20AND%20env%3Aproduction%20AND%20resource_name%3AKafka-PrintOfferConsumer%20operation_name%3Adeimos_consumer%20status%3Aerror&event=AwAAAZ2XwK7Yh_o_xQAAABhBWjJYd0xjTkFBQmhrMVFQMGRfQjNKczkAAABdMTE5ZDk3YzUtZTQ3MS00MmFlLWIwNGEtOGNhYmI3ZmZlNjIzLXN5bnRoZXRpYy10aW1lLTE3NzYzNjYwMDAwMDAwMDAwMDBjLTE3NzYzNjc4MjE5MzcwMDAwMDFvAAEBcg&fromUser=true&graphType=waterfall&historicalData=true&index=&panelFrom=1776366000000&panelTo=1776367800000&panelType=trace&refresh_mode=paused&shouldShowLegend=true&spanID=1835490373513295316&traceID=4439143481304951882&traceQuery=&view=spans&from_ts=1776272065872&to_ts=1776368241000&live=false), noticed that 5-error pattern in the trace with the first error being the deadlock and the final error that got surfaced being the Fk error.

## Fix

BatchRecord captures `primary_key_preset` at construction time. `fill_primary_keys!`'s guard short-circuits on that flag instead of on the live record[primary_col] value, since it can be stale. The existing optimization for user-supplied PKs is preserved, and retries ALWAYS get a fresh id lookup.

## How Has This Been Tested?

- [x] Rspec
- [x] Using item-platform PrintOfferConsumer (where the bug was originally found)

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+- Fix: `BatchRecordList#fill_primary_keys!` no longer short-circuits on stale in-memory primary keys left over from a rolled-back `DeadlockRetry` attempt, preventing `ActiveRecord::InvalidForeignKey` during association imports.
+
 ## 2.5.1 - 2026-03-18
 
 - Fix: Outbox producer would swallow ActiveRecord errors except for uniqueness.
--- a/lib/deimos/active_record_consume/batch_record.rb
+++ b/lib/deimos/active_record_consume/batch_record.rb
@@ -16,6 +16,8 @@ module Deimos
       attr_accessor :bulk_import_id
       # @return [String] The column name to use for bulk IDs - defaults to `bulk_import_id`.
       attr_accessor :bulk_import_column
+      # @return [Boolean] true if the primary key was supplied in the input attributes,
+      attr_accessor :primary_key_preset
 
       delegate :valid?, :errors, :send, :attributes, to: :record
 
@@ -32,6 +34,7 @@ module Deimos
         end
         attributes = attributes.with_indifferent_access
         self.record = klass.new(attributes.slice(*klass.column_names))
+        self.primary_key_preset = !self.record[klass.primary_key].nil?
         assoc_keys = attributes.keys.select { |k| klass.reflect_on_association(k) }
         # a hash with just the association keys, removing all actual column information.
         self.associations = attributes.slice(*assoc_keys)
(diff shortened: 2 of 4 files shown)


.DS_Store
dorner, line 1: Don't think this was supposed to be committed :)

dorner (changes requested):
PR looks good - just one accidental commit needs to be removed

.DS_Store
josephyang0519: oops. Fixed :P

dorner (approved):

PR #258, opened 2026-06-29
Extract Avro class generation to avro-gen-ruby

dorner:
This extracts the functionality to generate schema classes out of Deimos and into [avro-gen-ruby](https://github.com/flipp-oss/avro-gen-ruby/commits/main/). All existing functionality should remain the same.

This commit wires up the existing constants `Deimos::SchemaClasses::X` to point to `AvroGen::SchemaClasses::X` and emits deprecation warnings. The schema-class-related configuration remains supported and delegates to AvroGen configuration.

AvroGen ships with a task `avro_gen:upgrade` which will automatically find/replace the old constants.

Other than the deprecation warnings, no other effects should be noticeable.


lib/deimos/config/configuration.rb
```diff
@@ -15,6 +15,7 @@ module Deimos
       config.deserializers[:payload].try(:reset_backend)
       config.deserializers[:key].try(:reset_backend)
     end
+    Deimos::SchemaClass.sync_config!
```
harsha-flipp, line 18: I noticed that we are converting `Deimos.config` to `AvroGen.config`. I don't see a path here where settings can be deprecated or removed. Is that intended for a future release of the gem?

harsha-flipp (dismissed):

lib/deimos/config/configuration.rb
dorner: Nope - I intentionally wanted to continue to allow all configuration to happen in the same Deimos configuration block. Requiring multiple configuration blocks is unnecessary overhead for teams - they can just keep configuring it in the same place.

harsha-flipp (dismissed):

